PR #1, opened 2016-12-04
[Template] Bundler != Ember

segiddins:


indirect:
oh ffs, I somehow checked in the original file without my edits >_> thanks

PR #4, opened 2016-12-07
RFC for bundle add feature.

hmistry:
First stab at filling out our new RFC template for `bundle add` feature.

Look forward to seeing your inputs.

CC: @indirect, @segiddins 

[Rendered RFC text](https://github.com/hmistry/rfcs/blob/bundle-add/text/0000-bundle-add.md)

### DIFF
--- a/text/0000-bundle-add.md
+++ b/text/0000-bundle-add.md
@@ -14,7 +14,7 @@ The main use case is to deploy scripts and/or improve user productivity by addin
 
 The question is how far to extend this feature, is there a need to support all gem options like version, source, and group. And whether or not it should support multiple gems with their respective options.
 
-This command overlaps in function with the current `bundle inject GEM` command.
+This command overlaps some functionally with the current `bundle inject GEM` command.
 
 # Detailed design
 
@@ -33,10 +33,9 @@ This is the base use case. The following steps will occur when the command is ex
 1. It will load the current Gemfile and resolve the new GEM with it.
 2. If it fails to resolve correctly or is an invalid GEM, it will error out.
 3. If it resolves successfully, it will install the latest version of the GEM.
-4. It will add the GEM to the Gemfile with a comment and place a conservative version update requirement e.g.
+4. It will add the GEM to the Gemfile and place a conservative version update requirement* e.g.
 ```ruby
   # Gemfile
-  # Added at 2016-12-07 00:21:12 -0800 by username:
   gem "GEM", "~> 1.0.0"
 ```
 
@@ -45,15 +44,21 @@ This is the base use case. The following steps will occur when the command is ex
 ```ruby
 # First priority
 bundle add "GEM" [--version VERSION]
+bundle add "GEM" [-v VERSION]
 
 # Possible full implementation
 bundle add "GEM" [--version VERSION] [--source SOURCE] [--group GROUP]
+bundle add "GEM" [-v VERSION] [-s SOURCE] [-g GROUP]
 
 #examples
 bundle add "rails" --version "~> 5.0.0"
 bundle add "rails" --version ">= 5.0, < 5.1"
 bundle add "rails" --version "~> 5.0.0" --group "development, test"
 bundle add "rails" --version "~> 5.0.0" --source "https://gems.example.com" --group "development"
+bundle add "rails" -v "~> 5.0.0"
+bundle add "rails" -v ">= 5.0, < 5.1"
+bundle add "rails" -v "~> 5.0.0" -g "development, test"
+bundle add "rails" -v "~> 5.0.0" -s "https://gems.example.com" -g "development"
 ```
 #### Description
 This extends the base case allowing a user to specify optional version, source, and group.
@@ -62,22 +67,22 @@ The following steps will occur when the command is executed:
 1. It will load the current Gemfile and resolve the new GEM with it.
 2. If it fails to resolve correctly or is an invalid GEM/options, it will error out.
 3. If it resolves successfully, it will install the specified or latest version of the GEM.
-4. It will add the GEM to the Gemfile with a comment and the specified options. If the version is ">=0", then it will substitute it with a conservative version update requirement.
+4. It will add the GEM to the Gemfile and the specified options.
+
 ```ruby
   # Gemfile
-  # Added at 2016-12-07 00:21:12 -0800 by username:
-  gem "GEM" --version "~> 5.0.0"
-  gem "GEM" --version ">= 5.0", "< 5.1"
-  gem "GEM" --version "~> 5.0.0", group: ["development", "test"]
-  gem "GEM" --version "~> 5.0.0", source: "https://gems.example.com", group: "development"
+  gem "GEM", "~> 5.0.0"
+  gem "GEM", ">= 5.0", "< 5.1"
+  gem "GEM", "~> 5.0.0", group: ["development", "test"]
+  gem "GEM", "~> 5.0.0", source: "https://gems.example.com", group: "development"
 ```
 ## 3. add multiple gems
 #### Proposed command
 ```ruby
-bundle add "GEM1", "GEM2", "GEM3"
+bundle add "GEM1" "GEM2" "GEM3"
 
 #examples
-bundle add "rails", "devise", "simple_form"
+bundle add "rails" "devise" "simple_form"
 ```
 #### Description
 This extends the base case allowing multiple gems to be added.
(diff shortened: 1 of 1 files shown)


text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+```
+#### Description
+This is the base use case. The following steps will occur when the command is executed:
+1. It will load the current Gemfile and resolve the new GEM with it.
+2. If it fails to resolve correctly or is an invalid GEM, it will error out.
+3. If it resolves successfully, it will install the latest version of the GEM.
+4. It will add the GEM to the Gemfile with a comment and place a conservative version update requirement e.g.
+```ruby
+  # Gemfile
+  # Added at 2016-12-07 00:21:12 -0800 by username:
```
indirect: let's not have `add` include this comment. to start with, we can add the new line to the bottom of the gemfile. (eventually, I would like to put the gem in alphabetical order, assuming that the default gems in the gemfile are already sorted alphabetically)

text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+  gem "GEM", "~> 1.0.0"
```
indirect: the version that is chosen depends on the version of the gem that is available. for example, if the version chosen by the resolver is < 1.0, the constraint should have three places e.g. version `0.12.3` becomes `~> 0.12.3`. If the resolved version is > 1.0, the constraint should only have two places e.g. version `1.2.3` becomes `~> 1.2`.

text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+2. If it fails to resolve correctly or is an invalid GEM/options, it will error out.
+3. If it resolves successfully, it will install the specified or latest version of the GEM.
+4. It will add the GEM to the Gemfile with a comment and the specified options. If the version is ">=0", then it will substitute it with a conservative version update requirement.
+```ruby
+  # Gemfile
+  # Added at 2016-12-07 00:21:12 -0800 by username:
+  gem "GEM" --version "~> 5.0.0"
+  gem "GEM" --version ">= 5.0", "< 5.1"
+  gem "GEM" --version "~> 5.0.0", group: ["development", "test"]
+  gem "GEM" --version "~> 5.0.0", source: "https://gems.example.com", group: "development"
```
indirect: I think the `--version` needs to be removed from these examples?

text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+bundle add "rails" --version "~> 5.0.0" --source "https://gems.example.com" --group "development"
+```
+#### Description
+This extends the base case allowing a user to specify optional version, source, and group.
+
+The following steps will occur when the command is executed:
+1. It will load the current Gemfile and resolve the new GEM with it.
+2. If it fails to resolve correctly or is an invalid GEM/options, it will error out.
+3. If it resolves successfully, it will install the specified or latest version of the GEM.
+4. It will add the GEM to the Gemfile with a comment and the specified options. If the version is ">=0", then it will substitute it with a conservative version update requirement.
```
indirect: I'm not sure what `>= 0` means in this context?

text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+  # Added at 2016-12-07 00:21:12 -0800 by username:
+  gem "GEM" --version "~> 5.0.0"
+  gem "GEM" --version ">= 5.0", "< 5.1"
+  gem "GEM" --version "~> 5.0.0", group: ["development", "test"]
+  gem "GEM" --version "~> 5.0.0", source: "https://gems.example.com", group: "development"
+```
+## 3. add multiple gems
+#### Proposed command
+```ruby
+bundle add "GEM1", "GEM2", "GEM3"
```
indirect: we should not use commas to separate arguments to a command... no other commands do this

text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+  gem "GEM1", "~> 1.0.0"
+  gem "GEM2", "~> 1.0.0"
+  gem "GEM3", "~> 1.0.0"
+```
+
+## 4. add multiple gems with options
+#### Proposed command
+```ruby
+# TBD
+bundle add "MULTIPLE_GEMS + OPTIONS ???"
```
indirect: I think supporting options and multiple gems at the same time is probably too much. We can punt on this option and not plan to implement it for now.

text/0000-bundle-add.md
indirect: adding specific versions or version constraints should be allowed here using the same syntax as `gem install`: for example, `bundle add GEM1:2.0 GEM2:~>1.2 GEM3:>=4.0`.

text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+
+This will overlap functionality with the `bundle inject` command but it's more descriptive as it fits a users mental model better (add something vs inject something).
+
+# Alternatives
+
+*(What other designs have been considered? What is the impact of not doing this?)*
+
+# Unresolved questions
+
+1. Of the 4 categories above, which ones will be most commonly used to prioritize development? The current plan is first get category 1 (single gem) and category 2 (single gem with version option only) working.
```
indirect: I think it is fine to merge a PR that only supports category 1+2 to start, but it shouldn't be very much harder to expand that to include category 3. up to the implementor whether they happen all at once or category 3 is a follow-up PR.

text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+2. For category 2 (add a single gem with options), there are other cli format's to consider. However, I think these break consistency with bundler's cli and can present some confusion to users. Some alternatives being considered are:
+```ruby
+bundle add "GEM" [VERSION] [--source SOURCE] [--group GROUP]
+bundle add "GEM" [-v VERSION] [--source SOURCE] [--group GROUP]
+
+#examples
+bundle add "rails" "~> 5.0.0" --group "production"
+bundle add "rails" -v "~> 5.0.0" --source "https://gems.example.com" --group "development"
```
indirect: If we have `--version`, it should support `-v` as well. Same for `--source` and `-s`, and `--group` and `-g`.

text/0000-bundle-add.md
```diff
@@ -0,0 +1,132 @@
...
+```
+3. Will this eventually replace `bundle inject`? If so, what will be the deprecation plan?
```
indirect: No, it will not replace inject—inject is designed to be run by a completely automated tool with no human interaction or input, and it seems fine to leave that around for now.

text/0000-bundle-add.md
hmistry: Ignore it as it doesn't make sense. It should be what the user specified.

text/0000-bundle-add.md
hmistry: I think the bigger question for me is whether we break consistency with the rest of bundler's syntax and start incorporating short form options (`-v`, `-s`)? 

To me, breaking consistency leads to confusion kind of like what rails went through where beginners had to figure out which commands runs with rails and which runs with rake... now that they've fixed it, it's beautiful to use even for the long time rails devs.

My thought on this is if I introduced short form options here, then I'd plan on propagating it to all the other command options... something to consider.

text/0000-bundle-add.md
hmistry: I can implement the version constraints in that syntax, it's not a problem if that's really what you want. I find that syntax to not be very readable and thus can be prone to mistakes. It also won't be very scalable to adding other options in future like source, platform, etc so this will be just for gem + version. 

Regarding readability, I find the colon is not a good visual separator. It looks ok in your comment because the gems are in all caps. Here's an example to showcase my point, compare the following based on visual readability in your terminal:
```
# showcasing only, not suggesting to use these particular alternatives
bundle add rails:5.0.0.1 devise:~>4.2.1 simple_form:>=4.0
bundle add rails/5.0.0.1 devise/~>4.2.1 simple_form/>=4.0
bundle add rails[5.0.0.1] devise[~>4.2.1] simple_form[>=4.0]
```

segiddins:
Should we merge this? @hmistry @indirect @colby-swandale

hmistry:
@segiddins I think so. I updated it with @indirect previous feedback. We can always modify it if there are future changes.

@indirect the other day I saw where I got the idea that `bundle add` would replace `bundle inject`, it was in https://github.com/bundler/bundler/issues/4855. This can be decided later and doesn't prevent merging the RFC.

PR #5, opened 2017-03-04
Add `bundle report-issue`

colby-swandale:
This is a copy of @indirect's proposal from [bundler/bunder#4925](https://github.com/bundler/bundler/issues/4925)

### DIFF
--- a/text/0002-bundler-report-issue.md
+++ b/text/0002-bundler-report-issue.md
@@ -1,0 +1,61 @@
+- Start Date: 04/03/2017
+- RFC PR: 
+- Bundler Issue: 
+
+#### Note: This was taken from @indirect at [bundler/bundler#4925](https://github.com/bundler/bundler/issues/4925)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


hmistry:
[Nice rendered view of above RFC](https://github.com/colby-swandale/rfcs/blob/786dc30d9479fb9c7b13ea45250c26d0a535f95f/text/0002-bundler-report-issue.md)

hmistry:
I really like the overall concept. There's a lot of info collected. A couple of knee-jerk reactions are:
1. Since we know the type of error, can we only collect what's necessary? In other words, don't collect anything that's not remotely needed at all.
2. I presume there can be privacy and security concerns of the user. For example: the user paths in the shell can give out more clues about a github user, i.e. a different internet username. For companies, this can be a potential leak of their entire stack for outsiders to see - thinking of Gemfile.lock.
3. Will there be an interface for a user to preview and edit/approve before submission? As a user, I want this control.

colby-swandale:
Quick thought: Don't use Gists, unlike issues, gists can be deleted by the user as well when the user deletes their account.

indirect:

PR #10, opened 2018-02-10
Improvements to `bundle add`, plus `bundle remove`

shime:
Adding the RFC for `bundle add` improvements and `bundle remove` command, as described in [shime/bun#3](https://github.com/shime/bun/issues/3).

Looking forward to your input!

[Rendered RFC text](https://github.com/shime/rfcs/blob/bundle-add-improvements/text/0002-bundle-add-and-remove.md)

### DIFF
--- a/text/0002-bundle-add-and-remove.md
+++ b/text/0002-bundle-add-and-remove.md
@@ -100,7 +100,7 @@ It appends gems to the end of the group block if the group already exists. It al
 #### Proposed command
 
 ```shell
-bundle add sinatra:2.0 minitest:~>5.11
+bundle add 'sinatra:2.0' 'minitest:~>5.11'
 ```
 
 ### Description
@@ -181,7 +181,6 @@ end
 
 ```shell
 bundle remove minitest minitest-reporters
-bundle rm minitest minitest-reporters
 ```
 
 ```ruby
@@ -232,7 +231,7 @@ bundle add minitest sinatra
 Prints out and exits with error status code:
 
 ```shell
-Gem "minitest" already present in the Gemfile. Aborting.
+Gem "minitest" already present in the Gemfile. Skipping.
 ```
 
 ### 7.2. removing gem when it's not present
@@ -243,13 +242,13 @@ source "https://rubygems.org"
 ```
 
 ```
-bundle rm minitest sinatra
+bundle remove minitest sinatra
 ```
 
 Prints out and exits with error status code:
 
 ```
-Gem "minitest" couldn't be found in the Gemfile. Aborting.
+Gem "minitest" couldn't be found in the Gemfile. Skipping.
 ```
 
 
--- a/text/0002-bundle-add-and-remove.md
+++ b/text/0002-bundle-add-and-remove.md
@@ -1,0 +1,275 @@
+- Feature Name: bundle\_add\_and\_remove\_improvements
+- Start Date: 2018-02-10
+- RFC PR: (leave this empty)
+- Bundler Issue: (leave this empty)
+
+# Summary
+
+This feature adds improvements for `bundle add` command and adds
+the opposite command for removing gems called `bundle remove`.
+
+The purpose of this RFC is to get feedback on use cases, CLI format/syntax
+and feature options.
+
+# Motivation
+
+The main use case is improving user productivity. 
+
+This is done by adding gem(s) to the Gemfile with the latest version in the range specified from the command line. User saves time since they don't have to search Rubygems for the latest version. 
+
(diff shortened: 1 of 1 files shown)


text/0002-bundle-add-and-remove.md
```diff
@@ -0,0 +1,275 @@
...
+```ruby
+bundle add minitest --group test
+```
+
+```ruby
+# Gemfile after
+source "https://rubygems.org"
+
+group :test do
+  gem "minitest", "~> 5.11"
```
colby-swandale: I don't agree with adding a version constraint when it wasn't specified by the user. The version constraint should be defined  by the `--version` option that `bundle add` has already.

shime: `bundle add` already works this way if you don't specify the version. How should it behave if the version is not specified?

colby-swandale: ah ok, didn't know that. nevermind then

text/0002-bundle-add-and-remove.md
```diff
@@ -0,0 +1,275 @@
...
+```
+
+```ruby
+# Gemfile after
+source "https://rubygems.org"
+```
+
+### Description
+
+The `remove` command removes gems from the Gemfile, along with their corresponding group block if it remains empty after the deletion. By default it runs `bundle install` afterwards, which can also be skipped with the `--skip-install` flag.
```
segiddins: It is `remove` or `rm`? I'm in favor of `remove`, for symmetry with `add`

text/0002-bundle-add-and-remove.md
hmistry: @colby-swandale FYI, this was @indirect default requirement in the original `add` command. He wanted conservative versioning to be set as it's the safest and can be overridden if the user added a specific version.

text/0002-bundle-add-and-remove.md
hmistry: I too am in favour of `remove` for same symmetry reason and also most bundler commands are whole words, not abbreviations/shortened versions. User will therefore expect whole words as commands.

text/0002-bundle-add-and-remove.md
```diff
@@ -0,0 +1,275 @@
...
+```
+
+```
+bundle add minitest sinatra
+```
+
+Prints out and exits with error status code:
+
+```shell
+Gem "minitest" already present in the Gemfile. Aborting.
```
hmistry: I'd suggest changing `aborting` to `skipping`. Aborting sounds very serious and I wouldn't want it to abort in multi gem additions. I prefer it skips and adds remaining gems

text/0002-bundle-add-and-remove.md
```diff
@@ -0,0 +1,275 @@
...
+
+### 7.2. removing gem when it's not present
+
+```ruby
+# Gemfile
+source "https://rubygems.org"
+```
+
+```
+bundle rm minitest sinatra
```
hmistry: `bundle remove` if all agrees

text/0002-bundle-add-and-remove.md
```diff
@@ -0,0 +1,275 @@
...
+```
+
+Prints out and exits with error status code:
+
+```
+Gem "minitest" couldn't be found in the Gemfile. Aborting.
```
hmistry: Same comment as before `skipping` over `aborting`.

text/0002-bundle-add-and-remove.md
```diff
@@ -0,0 +1,275 @@
...
+which are more natural for users.
+
+It appends gems to the end of the group block if the group already exists. It also supports adding to multiple groups.
+
+## 2. adding multiple gems with options
+
+#### Proposed command
+
+```shell
+bundle add sinatra:2.0 minitest:~>5.11
```
booch: You're going to need quotes around the 2nd gem there, or the shell will interpret the `>` as output redirection. I'd recommend putting single quotes around all the gem names/versions in shell commands you show in the docs.

shime:
Thanks for your input, everyone! I've made the updates to the RFC document. Regarding @segiddins's comment - I don't think we can do much when it comes to arbitrary Ruby code inside a Gemfile. 

These commands would operate on the text in the Gemfile, they would not execute any Ruby code inside it. So for removing, it would report failure to do so if a gem is not found in a Gemfile text. Same goes for adding, it would check if a gem is present in a Gemfile text and if not it would try to add it to according group.

indirect:
Accepting this PR, since implementation has already begun as part of GSoC 2018. 😄

PR #14, opened 2018-07-23
RFC for `change` command

agrim123:


text/0003-bundle-change.md
```diff
@@ -0,0 +1,107 @@
+- Feature Name: bundle\_change
+- Start Date: 2018-07-22
+- RFC PR: (leave this empty)
+- Bundler Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes a new bundle change command. The idea behind it is to make Gemfile editable via command line.
```
colby-swandale: It would be good to say what the command does in 1 sentence.

text/0003-bundle-change.md
```diff
@@ -0,0 +1,107 @@
...
+
+# Guide-level explanation
+
+At an abstract level the change command is a composite of remove and add command. the requested gem would be first removed and then added with the supplied options (which it supports).
```
colby-swandale: > the requested gem would be first removed and then added with the supplied options

This needs more explanation on what this does and why removing and then adding the gem is better than replacing it.

colby-swandale:
Does this command run `bundle install` once the change is made? Is this optional? What happens if the user runs `bundle change` but the gem cannot be installed?

How does it handle parsing the Gemifle? What happens if the gem is declared in a non-standard way in the Gemfile?

text/0003-bundle-change.md
agrim123: It was the initial idea to remove the line which depends on `remove` and then add it.

agrim123:
> Does this command run bundle install once the change is made?

Yes. Basically its the add command that does this.

> Is this optional?

Not yet but this can be made optional.

> What happens if the user runs bundle change but the gem cannot be installed?

Assuming that gem has been successfully removed and added but error occurs when add runs bundle install. I think add will handle that and throw appropriate error.

> How does it handle parsing the Gemifle? What happens if the gem is declared in a non-standard way in the Gemfile?

This is all dependent on remove command. If gem could not be found/removed or extra gems are removed then the remove command throws error and the Gemfile is reverted.

segiddins:
@agrim123 sounds good!

simi:
Given the custom `Gemfile.lock` and Ruby nature of `Gemfile` files, it is (at least in my eyes) super complex to manipulate with those files in safe way. I was thinking for long time to support alternative format which would fit majority of needs backed by YAML (for both `Gemfile` and `Gemfile.lock`).

This kind of alternative file would be super easy to modify in non-interactive way with `bundle add/remove/change/...` commands. On the other side it would not be able to evaluate Ruby in `Gemfile` anymore (which per my experience is not widely used, but could be still helpful).

Is it worth to create initial RFC for this kind of alternative `Gemfile`/`Gemfile.lock` syntax?

deivid-rodriguez:
I think `RuboCop` is pretty good at enforcing `Gemfile` styles, I think we should not mess with that. I also think `bundle change` is too hard and in general I don't want to add more commands that manipulate `Gemfiles` in this way.

If we had an alternative non dynamic YAML/JSON format, that would make this much easier indeed. But I'm not sure of the chances that someone steps in and implements something like that. So I'm not fully sure whether it's worth spending time on a RFC.

PR #18, opened 2019-04-23
RFC on merger with RubyGems

indirect:
It's happening!!! 🎉 

[Rendered](https://github.com/bundler/rfcs/blob/ffaa5d7b21d454ae8d528180cdc11b32120a2ccf/text/000-merge-with-rubygems.md)


jules2689:
I understand and appreciate the overall idea, but it does raise a bit of a question around competition. When rubygems and bundler are together, there will be less incentive and it will make less sense to create tools like gel, which has already spawned some healthy discussion.

Competition encourages the betterment of everything, and merging the 2 projects may inhibit that.

text/000-merge-with-rubygems.md
```diff
@@ -0,0 +1,100 @@
...
+From time to time, Bundler may introduce opt-in changes that require a certain minimum version of Bundler. In that case, it is possible to choose to upgrade an application to require the latest version of Bundler in order to use those features.
+
+However, Bundler will never force an application (or other users of an application) to upgrade to the latest Bundler version. The minimum required version of Bundler will always be opt-in, per application, and never be systemwide across an entire Ruby.
+
+## Reference-level explanation
+
+Implementing the merger will likely require multiple phases. The organizational merger is easy to start right away, while the code-level merger will require some research and experimentation beforehand.
+
+### Phase 1, organization merger (April 2019–May 2019)
+- Move GitHub repos from the `bundler/` org to the `rubygems/` org
```
colby-swandale: IIRC we would only move the `bundler/bundler` repo, everything else will be left alone.

text/000-merge-with-rubygems.md
```diff
@@ -0,0 +1,100 @@
...
+- Add all members of the Bundler maintainers team to the existing RubyGems maintainers team to create a new combined maintainers team
```
colby-swandale: Should be noted that this is only for the core team. The committer team from both projects will not be combined and will only have access to their respective project. Ie: Rubygems committers **will not** be given commit bit to bundler.

text/000-merge-with-rubygems.md
bronzdoc: The bundler org will disappear?

indirect: I think we would also want to move at least this RFCs repo and the Gemstash repo? I think it will be easiest to just move everything though, and leave a repo in bundler/ with a note. GitHub will set up redirects for the repos.

text/000-merge-with-rubygems.md
bronzdoc: > Add all members of the Bundler maintainers team to the existing RubyGems maintainers team to create a new combined maintainers team

This is good, it will make sure we all work towards the same goal.

indirect:
@jules2689 can you elaborate on what you think will be different? Today, Bundler and RubyGems both already depend on each other’s internals, and that didn’t stop Gel from happening.

jules2689:
Bundler relies on Rubygems internals, I knew that. But I didn't think Rubygems relied on Bundler in any way.

I'm not saying it's a blocker, but a potential downside. Having these so tightly coupled as to become one would kind of discourage competition IMO. Gel could happen because Rubygem's sole responsibility is to host gems. Bundler does all the fancy stuff with collecting gems, dependency resolution, etc. There is a clear line that isn't crossed, which means anyone can replace the latter half. When they are combined to one, that line is fuzzy and concepts blurred, such as the differentiator I used in #17 to determine that Bundler is a more appropriate place to put the error handling. Given that that line would be completely grey, a library maintainer for something like gel would not be able to rely as easily, I'd think, on just the Rubygems portion without fear of "this is literally only made for Bundler now, well it is bundler now". 

This is speculation alone, of course.

deivid-rodriguez:
>  Rubygem's sole responsibility is to host gems. Bundler does all the fancy stuff with collecting gems, dependency resolution, etc. 

This is not true. [rubygems.org](https://github.com/rubygems/rubygems.org) responsability is to host gems, and it's not going anywhere. The rubygems CLI does essentially the same thing as the bundler CLI, only in a more limited way.

jules2689:
> The rubygems CLI does essentially the same thing as the bundler CLI, only in a more limited way.

That is a fair point. I had conflated the two, and I think the distinction is more clear in my mind now. That's for bringing that up.

text/000-merge-with-rubygems.md
jules2689: It may make sense to clean things up first. https://github.com/bundler/bundler-slackin, for example, is no longer needed as Slack can make permanent invite links now (I've had a url redirect to one for another slack group - https://chat.devproductivity.io).

indirect:
Copied over from bundler/rfcs#17, here's a brief explanation of each project:

RubyGems.org: provides 1) a list of all gems, 2) gemspecs, and 3) the .gem files. There are a few different formats for that data, but that's the gist.

RubyGems: provides 1) the gem command, 2) a monkeypatch to require that knows how to find installed gems, 3) the Gem class and everything underneath that.

Bundler: provides 1) the bundle command, and 2) the bundler/setup Ruby file that monkeypatches require (and a lot of other parts of RubyGems) to limit it just to the single application bundle, and 3) the Bundle class and everything underneath that.

As you can maybe imagine, Bundler depends on lots of RubyGems internals, and in the last few years we have been starting to consolidate places where RubyGems and Bundler have two different classes to do the same thing with a single class that both of them use. For example, they both share the same dependency resolver, Molinillo, and they both share the same client for the newest API to get information about gems.

text/000-merge-with-rubygems.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Guide-level explanation
+
+** Before the code-level merge:**
+
+Bundler 3 is a stepping stone on the way to combining Bundler and RubyGems. You can install it by running `gem install bundler`. You will notice that there are some changes to how the `bundle` command works.
+
+We've been planning and working towards those changes for a long time, and we think you'll find Bundler easier and more straightforward to use once you've had a chance to get used to the changes.
+
+We've already started working on the next version of Bundler, which will be fully part of RubyGems. It will not need to be installed separately, and will always be available anywhere you have Ruby and RubyGems available.
```
segiddins: > It will not need to be installed separately

can you clarify: can it be updated separately? will ruby gems work without it?

segiddins: > anywhere you have Ruby and RubyGems available

is distro package manager un-bundling a concern here?

text/000-merge-with-rubygems.md
```diff
@@ -0,0 +1,100 @@
...
+### Phase 1, organization merger (April 2019–May 2019)
+- Move GitHub repos from the `bundler/` org to the `rubygems/` org
+- Add all members of the Bundler maintainers team to the existing RubyGems maintainers team to create a new combined maintainers team
+- Combine maintainer Slack workspaces and Slack channels
+- Combine security@, team@, and other emails
+- Combine READMEs, issue reporting guides, and other in-repo docs
+- Update documentation and website links to issues, pull requests, troubleshooting, developer docs, etc
+
+### Phase 2, research for codebase merger (June 2019–??? 2019)
+- Research and test a merge plan that replaces the Bundler submodule in RubyGems with the entire git history and all commits from the Bundler git repo
```
segiddins: I think I wrote a test script that would do this, I can try and did it up later

segiddins: How will we handle the different code styles / practices used by the two parts of the repo?

text/000-merge-with-rubygems.md
```diff
@@ -0,0 +1,100 @@
...
+- Create a CI task that runs run all of the RubyGems and Bundler tests together, covering the current RubyGems and Bundler build matrix (we can delete SO MANY MATRIX ENTRIES 🤩)
+- Update the Bundler release scripts and automation to release out of the RubyGems repo, and test it
+
+### Phase 3, codebase merger (??? 2019–??? 2020)
+- Use the plan from Phase 2 to merge the Bundler repo into the RubyGems repo in place of the submodule. It's ok if this needs a force push, but we want to avoid more than one force push.
+- Ensure that CI continues to pass with Bundler development moved into the RubyGems repo
+- Use GitHub's "Move Issue" beta feature to move active issues from the Bundler repo over to the RubyGems repo
+- Update the `rubygems/bundler` readme and "new issue" message to point to `rubygems/rubygems` for everything except bug fixes for old Bundler versions
+- Write and test a release script that automatically releases both RubyGems and Bundler together with one version number, one changelog, one tag, and one stable branch, as an artifact that can be installed via `gem install --system`
```
segiddins: how will rubygems handle different versions of the bundler gem being installed / attempted to be used? 

how will bundler handle attempts to be run with a different version of rubygems?

segiddins:
> they both share the same client for the newest API to get information about gems

I don't believe this is true yet, I don't think the compact index is supported by RubyGems

text/000-merge-with-rubygems.md
hsbt: > how will bundler handle attempts to be run with a different version of rubygems?

Is it in the test environment? The rubygems never work the multiple versions with the single ruby version in the real-world usage.

text/000-merge-with-rubygems.md
segiddins: https://gist.github.com/segiddins/7a846516bd15f6655120db5f91e11e42, found it, but if i'm reading it correctly the point of my script was to make it easy for people with pending contributions to move their PRs over to the rubygems repo

text/000-merge-with-rubygems.md
segiddins: I meant, what will happen if someone installs a bundler using a different version of rubygems than we intend to support

colby-swandale:
@indirect We also need to figure out what we'll what we want to do with bundler.io as well.

text/000-merge-with-rubygems.md
indirect: A script to help people move their pending PRs over sounds helpful, thanks!

text/000-merge-with-rubygems.md
indirect: Assuming we are able to work out the technical/UX concerns (I have a proposal that I will open as a separate RFC soon), it will not be possible to update Bundler separately from RubyGems. 

Distros that break up stdlib always set dependencies on the versions that ship together, so it shouldn't be any more broken than the usual stdlib splitting up already is.

text/000-merge-with-rubygems.md
indirect: Bundler and RubyGems fully combining means you will only be able to install one copy of both at the same time. To be clear that is a future, eventual goal: the immediate goal is to get them both shipping from the same repo with their own separate versions and separate packages and separate installation.

Later, after releasing both packages out of one repo is working, I am proposing that we plan to combine releases, with one single "RubyGems/Bundler" thing that can only be installed as one blob, and that can only have one version per Ruby.

indirect:
@segiddins good point on the compact index not being implemented inside RubyGems yet. Hopefully we can avoid having two copies of that functionality, and reuse one copy instead. 😬 

@colby-swandale so far, I had been imagining that bundler.io would continue to host the docs for the `bundle` command, but I am open to other possibilities.

text/000-merge-with-rubygems.md
indirect: > It may make sense to clean things up first. https://github.com/bundler/bundler-slackin, for example, is no longer needed as Slack can make permanent invite links now (I've had a url redirect to one for another slack group - https://chat.devproductivity.io).

Sadly, Slack does not offer permanent links—only "up to 90 day" links. So we're going to keep Slackin.

indirect:
Thanks for the feedback, everyone! Here we goooooooo......

PR #19, opened 2019-06-19
Working with stdlib gems

indirect:
[Rendered](https://github.com/rubygems/rfcs/blob/indirect/stdlib_gems/text/0000-stdlib-gems.md)

### DIFF
--- a/text/0000-stdlib-gems.md
+++ b/text/0000-stdlib-gems.md
@@ -1,0 +1,41 @@
+- Feature Name: stdlib-gems
+- Start Date: 2019-06-18
+- RFC PR: (leave this empty)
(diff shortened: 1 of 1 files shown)


voxik:
Have somebody considered to use something like [Modulation](https://github.com/digital-fabric/modulation)? That would make the bundling easier. In fact, that would allow to ship the packages unmodified as just different version of gem and load them for Bundler in its own namespace, leaving regular user to choose whatever version is required.

deivid-rodriguez:
@indirect We have actually been using this strategy for a while, so I think this is good to merge since it's something we already do.

simi (approved):

indirect:
👍🏻 will merge

PR #23, opened 2020-01-22
Add RFC proposal for adding bundle fund

gjtorikian:
This is an RFC for the suggestion proposed in https://github.com/rubygems/rfcs/issues/22: adding a `bundle fund` command which shows information about any gems requiring funding. I look forward to hearing any feedback!

### DIFF
--- a/text/0000-add-funding-uri.md
+++ b/text/0000-add-funding-uri.md
@@ -90,7 +90,7 @@ $ bundle info GEM
 	Funding:  GEM_FUNDING_PAGE
 ```
 
-If there is no `funding_page` attribute, no information will be displayed. If none of the gems have a `funding` like, the command should print a message out indicating that:
+If there is no `funding_page` attribute, no information will be displayed. If none of the gems have a `funding_page` defined, the command should print a message out indicating that:
 
 ```
 $ bundle fund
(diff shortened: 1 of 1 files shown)


olleolleolle:
Is there a specific reason why `funding_uri` is not a part of the [metadata](https://guides.rubygems.org/specification-reference/#metadata)?

gjtorikian:
> Is there a specific reason why `funding_uri` is not a part of the [metadata](https://guides.rubygems.org/specification-reference/#metadata)?

I never knew about this attribute before. It certainly makes sense to add there as well.

olleolleolle:
You can read more details about this "sandboxed, freeform" hash of URIs, in the Specification source code: https://github.com/rubygems/rubygems/blob/master/lib/rubygems/specification.rb#L365

gjtorikian:
Ok, I've made a quick change to the RFC to note that this key should go in the `metadata` attribute.

olleolleolle:
(Glad you liked it. Perhaps it reduces friction!)

gjtorikian:
I do believe so, especially since it matches https://github.com/rubygems/rubygems/pull/3060

gjtorikian:
Is there anything I can do here to move this along?

indirect:
I’m in favor of this, but I’d like to give the rest of the team a chance to weigh in. Let’s give it 3 more days and then accept if no new comments arrive. 👍🏻

text/0000-add-funding-uri.md
```diff
@@ -0,0 +1,141 @@
...
+
+This is a backwards-compatible, additive change to bundler and the gem specification.
+
+# Reference-level explanation
+
+Right now, there is a very similar "print a message" gem attribute by way of `post_install_message`. A crucial difference between `post_install_message` and `funding_uri` is that `post_install_message` ought to be preserved for essential information. Sometimes, open source maintainers abuse this attribute by printing their own non-essential messages, to the annoyance of users. To this end, bundler added an `ignore_messages` configuration option, which silenced `post_install_message`s on an individual gem or a global level.
+
+I do not believe that displaying a single message indicating gems which need funding reaches this level of annoyance. Therefore, I do not expect `ignore_messages` to have an impact on `bundle fund`, or the `X gems are looking for funding` message printout, nor should an opt-out flag or configuration should be considered at this time.
+
+`bundle init` should also not add a placeholder for `funding_uri`.
```
colby-swandale: Do you mean `bundle gem` here? `bundle init` only just generates a `Gemfile`

gjtorikian: I did!

colby-swandale:
This has been open to comment for enough time i think and we're in favor of the RFC. Accepted.

PR #24, opened 2020-02-06
Proposal for lockfile versioning

indirect:
This proposal tries to increase Bundler's expected backwards compatibility in future versions, while still allowing changes to the lockfile format if they are needed in the future.

[Rendered text](https://github.com/rubygems/rfcs/blob/indirect/lockfile_versions/text/0000-lockfile-versions.md)

### DIFF
--- a/text/0000-lockfile-versions.md
+++ b/text/0000-lockfile-versions.md
@@ -1,0 +1,83 @@
+# Version lockfiles, not Bundler
+
+- Feature Name: lockfile-versions
+- Start Date: 2020-01-30
+- RFC PR: (leave this empty)
+- Bundler Issue: (leave this empty)
+## Summary 
+
+Many years ago, the Bundler and RubyGems teams made an assumption that Bundler 1 and the future Bundler 2 would not both work with the same Gemfile.lock, to protect against possible future breaking changes. This assumption turned out to be wrong—Bundler 2 uses the same Gemfile and lockfile as Bundler 1. When Bundler 2 finally came out, some unfinished features and some bugs combined to break many applications and CIs that still depended on Bundler 1. For the technical details, read this [blog post about Bundler 2](https://eregon.me/blog/2020/01/13/a-migration-path-to-bundler2.html) by Benoit Daloze.
+
+To ensure that future versions of Bundler will not break existing applications, this RFC proposes adding explicit versions to the Bundler lockfile format. In this RFC, we will call the current lockfile format “lockfile v1” and the upcoming (unreleased) lockfile format “lockfile v2”. By versioning the lockfile, we hope that any developer will be able to upgrade to any new Bundler version at any time, and keep working on the same projects, *without* forcing other developers on those projects to upgrade.
+
+
+## Motivation
+
+The Bundler 2 release was confusing for a complicated set of reasons, including a partially-functional automatic version switcher, some bugs in Bundler, some bugs in RubyGems, and some confusing error messages.
+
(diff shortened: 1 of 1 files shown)


colby-swandale:
\cc @schneems

eregon:
This looks great to me, and I think it will provide a much better migration path for newer versions of Bundler. The Reference-level explanation is pretty close to the [suggestions](https://eregon.me/blog/2020/01/13/a-migration-path-to-bundler2.html#a-better-migration-path-for-bundler-2) in my blog post :smile:
Thank you for creating this RFC.

hmistry:
I like the idea of versioning lockfiles very much. I think it's a good way to decouple improving lockfiles and bundler versions while allowing backward compatibility. I would consider if Gemfile should also have versioning incase you'd like to make changes to it or have new features... food for thought.

With versioned lockfiles, I'm not sure I see a need for having bundler versions in the lockfile but maybe there is some use case or it doesn't hurt.

Normally I'd also think about how you'd migrate from one version to another but I don't see this being an issue here since bundler will parse the Gemfile and regenerate a new lockfile version. But let's say if this were Yaml files with user set fields, then migrating old Yaml files to a new version while transferring user data becomes an issue if it's not an easy 1 to 1 map (something I'm facing in another project)... again I don't see this being an issue here.

I think it would be good to specify an end of life plan for lockfile version support from the beginning to set expectations. You don't want to have the same issue 10 years from now having implementation complexities to support old versions like you went through with Bundler.

schneems:
Thanks a ton for this. The proposal matches closely what we discussed at RubyConf. I like it. 

Sorry for the delayed response, I'm giving this my 👍

schneems:
One suggestion: I would like if there is a canonical location that documents what version of bundler introduced what version of the lockfile. For instance "Bundler 2.0.0 introduced lockfile format v2". While presumably getting a lockfile vers error would lead the developer to upgrade to the latest bundler, if for some reason there is a bug, feature, or some other concern doing that then it would be helpful to not have to guess and check which bundler versions provide support for which lockfile formats. Documenting the support somewhere will be helpful for that case and possibly for debugging.

duckinator:
I'm definitely in favor of adding lockfile versioning. I'll go through your "unresolved questions" section and give my thoughts on some of them.


> What implementation option do we want to choose to support installing v1 and v2 lockfiles, as well as allow one-way upgrades from v1 to v2? Separate lockfile parsers per version? Version flags and lots of conditionals?

I think it's worth having the parser and such for each lockfile version be separate implementations.

Here's my reasoning for that:

1. It allows us to **rip out old cruft in new lockfile versions**. This is not something we could do easily if we try to make one parser support every version.
2. It allows us to **make more drastic changes, like completely changing the format**.
3. It allows us **move old parsers out to Bundler plugins** &mdash; as usage of a lockfile version drops, we can remove them from the primary codebase while still providing a method for people to use them if necessary.
4. If anything winds up being truly 100% equivalent between two lockfile versions, we can add a mechanism to allow them to reference an old version's functionality. This feels a lot more reasonable and maintainable, to me, than having every lockfile version use the same code by default and just have a pile of conditionals everywhere.


> How should we communicate with users about this? Bundler.io and RubyGems.org blog posts? Conference talks?

Something along these lines, perhaps:

- have a dedicated page discussing the overall topic on either the RubyGems or Bundler website (whichever is deemed more appropriate) and documenting bundler/rubygems versions support which lockfile versions, which links to separate pages for each lockfile version discussing any problems/limitations, new features, etc.
- on whichever website _doesn't_ have that page, add a page about it that is nothing but a short explanation and a link to the other one. (that way folks can find it even if they go to the "wrong" website!)

indirect:
Thanks for the discussion, everyone. Let’s do this!

Issue #31, opened 2021-01-07
Discussion: Namespaced Gems?


mullermp:
I wanted to open a discussion about supporting gem namespaces. Between this repo and the RubyGems.org repo, I wasn't sure where to start this discussion. If this was opened in the wrong place, I apologize, and feel free to close it.

Regarding the issue, though: currently, anyone can go and claim any gem they want, and it is difficult for organizations with a large number of gems (such as rails, aws, etc) to reserve rights to naming. It's not only confusing to customers, but can also be a concern for security if a malicious actor claims a similarly named gem and labels it as official. Other package managers (such as NPM's scopes) have support for namespaces.

My thoughts would be to use a `/` or `::` for namespacing, such as `gem install aws::s3` or `gem install rails/actionwhatever`

mullermp:
I'm sorry for the radio silence on this issue. I've been caught up with other work related things. This is still a useful feature with large impact that I currently don't have the bandwidth to drive. If someone wants to own this, by all means please do.

Fryguy:
@indirect What was the reason for closing?  Is it just because a formal RFC wasn't created yet?

indirect:
This is a discussion ticket, and the discussion ended in March. What would the reason be for leaving this ticket open?

mullermp:
Can we re-open the discussion in the ruby gem RFC repo? I haven't had time to write an actual RFC around this, but overall I think it's a useful feature/design/discussion to be having.

mullermp:
Proposal is here: https://github.com/rubygems/rfcs/pull/40

PR #33, opened 2021-09-09
Gem ownership custodians and a process for dealing with abandoned gems.

ioquatix:
cc @sonalkr132

### DIFF
--- a/text/0007-ownership-custodians.md
+++ b/text/0007-ownership-custodians.md
@@ -18,7 +18,7 @@ I propose we introduce a new clearly documented and audited process to deal with
 
 ### Gem Adoption Process
 
-- Given that a user makes an **ownership request** on a specific gem, after a certain time period (e.g. 7 days) they can request a gem custodian review.
+- Given that a user makes an **ownership request** on a specific gem, after a certain grace period (e.g. 7 days) they can request a gem custodian review.
 - The gem custodian can review all details of the user making the request, including all other gem **ownership request**s they have made, and the details of the current gem **ownership request**.
 - Within certain guidelines, they have the ability to approve a gem **ownership request** without the express consent of the current owner.
 	- The gem must meet the general criteria of being abandoned.
@@ -30,7 +30,7 @@ I propose we introduce a new clearly documented and audited process to deal with
 	- Gems which do not meet the requirements for being abandoned require the approval of 3 gem custodians.
 	- Any gem custodian can veto a gem **ownership request**.
 
-If this process fails because the current owner rejected the transfer, the **ownership request** can be revisited after a period of 12 months. If the gem then meets the criteria of being abandoned, this process may be repeated, and the gem custodian should consider this history in their decision making process.
+If this process fails because the current owner rejected the transfer, the **ownership request** owner can re-request a gem custodian after a period of 12 months. The gem custodian should consider this history in their decision making process.
 
 ### Abandoned Gem Criteria
 
--- a/text/0007-ownership-custodians.md
+++ b/text/0007-ownership-custodians.md
@@ -1,16 +1,22 @@
 # Ownership Custodians
 
-The health of <https://RubyGems.org> is important for the community and there are a large number of factors which decide this. Having a large number of essentially abandoned gems contributes negatively towards this and we would like to continue improving the mechanisms for dealing with this problem.
+The health of <https://rubygems.org> is important for the community. Having a large number of essentially abandoned gems contributes negatively towards the usability and impression of RubyGems, and we would like to improve the mechanisms for dealing with this problem.
 
-We propose to introduce gem custodians and a process for approving **ownership request**s where the original owner is unresponsive or uncooperative. This would be an accountable process to deal with abandoned gem namespace.
+Developers who want to release new Ruby libraries are often hampered by the lack of good available gem names. There is no formal process for reusing gem names which are effectively abandoned. Informal processes often break down due to lack of trust or uncontactable gem owners. This can create friction for developers as they then have to consider alternative names which are less ideal.
+
+For users who search RubyGems to find relevant projects, they are often presented with many options which have not been updated for many years. This presents a bad impression of the Ruby community as "unmaintained". This problem will only get worse over time if no effort is made to address the underlying reasons.
+
+We propose to introduce gem custodians and a process for approving ownership requests where the original owner is unresponsive or uncooperative. This would be an accountable process to deal with abandoned gem namespace and enable valuable gem namespace, which is effectively abandoned, to be reused in a way which is ultimately positive for the Ruby community.
 
 ## Motivation
 
-There are several cases where the current gem **ownership request** process can be inadequate.
+There are several cases where the current gem ownership request process can be inadequate.
 
-- **Unresponsive/Deceased Owners**: Sadly, from time to time, gem owners will pass on. In this case, it is not expected that there would be any response from the owner. We need to consider both popular and unpopular gems. Appropriate mitigating factors (like multiple authors) may not be in place when the unfortunate situation occurs.
+- **Deceased Owners**: Sadly, from time to time, gem owners will pass on. In this case, it is not expected that there would be any response from the owner. We need to consider both popular and unpopular gems. Appropriate mitigating factors (like multiple authors) may not be in place when the unfortunate situation occurs.
 
-- **Uncooperative Owners**: Unfortunately, even if contactable, some gem owners are uncooperative. This is understandable as the **ownership request** might come from someone who is unknown to the original owner, and the various issues surrounding gem ownership (including hijacking important gems, etc).
+- **Unresponsive Owners**: Not all gem owners provide their contact details, nor is there any guarantee that their contact details will work. This makes it difficult for informal gem ownership trasferral to occur.
+
+- **Uncooperative Owners**: Unfortunately, even if contactable, some gem owners are uncooperative. This is understandable as the ownership request might come from someone who is unknown to the original owner, and the various issues surrounding gem ownership (including hijacking important gems, etc).
 
 I propose we introduce a new clearly documented and audited process to deal with these cases. Some members of the RubyGems community are nominated as custodians (could include members of Ruby core team for example, admins on RubyGems, etc). Some pre-requisites like experience, activity, etc may be considered. Those members have the ability to go through a gem adoption process either for themselves or on behalf of another member of the community.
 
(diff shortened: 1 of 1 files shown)


ioquatix:
As part of this, we might also want to consider the inverse case: blocking ownership transfers that look suspicious.

https://mjtsai.com/blog/2018/11/27/popular-npm-package-compromised/

In some cases, authors of popular packages might transfer ownership in a way which produces poor outcomes for the community. We may want to introduce a gem custodian review of gem ownership transfers which could potentially fall into this category.

mensfeld:
Couple remarks here:

> Unresponsive Owners: Not all gem owners provide their contact details, nor is there any guarantee that their contact details will work. This makes it difficult for informal gem ownership trasferral to occur.

Not every owner wants to be reached out to. I've stumbled across this and well, it's their right to do so.

> Uncooperative Owners: Unfortunately, even if contactable, some gem owners are uncooperative. This is understandable as the ownership request might come from someone who is unknown to the original owner, and the various issues surrounding gem ownership (including hijacking important gems, etc).

"Forcefull" transfer of ownership is always a high risk procedure. This can be a potential security vulnerability by itself.

> The gem custodian can attempt to communicate with the current owner on behalf of the ownership request. 

There should be a way to "opt out" from this type of communication either per account of per gem.

> A gem is considered "in use" if the gem itself, or any of its direct reverse dependencies (version specific) have more than 10,000 downloads in the past year.

There are semi-private gems that will never reach it. Download count is not the best to assess usage.

> A gem is considered "maintained" if the author has logged into https://RubyGems.org within the past 12 months.

Also insufficient. I've seen gems that did not have releases for 7 years but have small updates on Github.

> A gem is considered to be "valid" if the homepage and other related metadata is valid (e.g. URLs return relevant pages).

There are literally hundreds of gems without home page, putting rubygems as a home page or with invalid github link (typo, account, etc)

> A gem is considered to be "working" if a user can check out and run the unit tests on a supported, non-EOL Ruby implementation.

This would mean we should encourage ppl to add tests into the releases, while for majority of cases it does not make any sense. Also there are many gems that have external deps (DBs, queues, etc) and won't "just run".

> A gem is considered to be "stable" if it has a published release >= 1.0.0.

Also not a standard.

> A gem is considered "abandoned" if the the most recent release of the gem exceeds the "adoption threshold".

Having so clear definition of that nature means, I could farm this data, create many RG accounts and slowly start taking over many packages to build up a network of packages I could then use.


I would love to come up with a process here, but I feel that we need to have a deeper discussion. 

I would start from asking: how often does this happen? I recall two cases for at least a year now. I do agree we need to have policies around this as well as a systematic way but I'm not sure if it should be "popularity" based.

ioquatix:
I don’t know the number off the top of my head but even the “async” gem was due to private ownership transfer. If it wasn’t for someone’s generosity to give me a name that made sense I might have just given up. A large number of the gems I created and maintain are due to ownership transfer. I feel that because of the lack of name availability, and the lack of transparent process, many developers could give up due to the friction. That’s just my guess though - how would you measure it? The ownership transfer process is not very easy for either party.

I currently have sent out two email messages to gems which have not been updated for many years, to try and reuse them. But after one week did not hear any reply. I don’t think it’s fair to use a shared resource and refuse to have reasonable communications. Even DNS has requirements about contact details which you have to update on a yearly basis.

I agree all your points but that doesn’t mean we don’t need some more transparent process. I think the motivations I provided are real problems. You’d definitely be in a more knowledgeable position to suggest some kind of solution.

Regarding people hosting small gems which are effectively private, is Rubygems, which is a shared community resource, really the right place for it?

I would even argue that taken to its extreme, Rubygems is not sustainable. Over time it seems like it would just get more and more saturated with unmaintained gems. Without any “corrective pressure” the problems that I’ve outlined seem like they will only be worse in the future.

I don’t know what the solution is but I feel a transparent process for dealing with the situations I’ve outlined would be helpful. Maybe having expectations around communication (e.g. DNS) would be another.

Ultimately I would like gems which are effectively unmaintained and unused to be open to the community reuse more easily . But maybe this has unacceptable security risks. That being said, maybe we can define a process where the security risk is acceptable and also allows us to avoid these problems, i.e. the bar for any kind of “forced transfer” should be proportional to the perceived risk, which is what I’ve tried to capture in my proposal.

ioquatix:
Thanks to @nateberkopec who linked me to https://docs.npmjs.com/cli/v6/using-npm/disputes which outlines how NPM solves this problem. Well, they are certainly more direct in their approach.

Fryguy:
It feels like a lot of the points that were made can be resolved by just having a simple namespace solution.  Right now, we are all effectively putting gems into a global namespace.  With a namespace, you can use the same name in your own namespace, and then there's not necessarily a need to "transfer" unless there is an official namespace.  However, in that case of an official namespace, I'd expect that there is usually some sort of team in place that can take over.  Also, with a namespace, you solve the "private gems" problem.

Of course that doesn't solve all of the problems, particularly with dispute resolution, but... baby steps.

Now, _how_ to introduce a namespace is the interesting part.  I feel like there could be a transition where the current set of gems are moved into a "global" namespace; `gem` calls to a non-namespaced version automatically go to the global namespace; perhaps new gems can only go into namespaces; maybe have a way for gem maintainers of global gems to move them into a namespace with a redirect from the global namespace.  It will probably take years to fully transition, and then perhaps eventually unmaintained global gems can ultimately move to an "abandoned" namespace.

text/0007-ownership-custodians.md
```diff
@@ -0,0 +1,113 @@
...
+
+Developers who want to release new Ruby libraries are often hampered by the lack of good available gem names. There is no formal process for reusing gem names which are effectively abandoned. Informal processes often break down due to lack of trust or uncontactable gem owners. This can create friction for developers as they then have to consider alternative names which are less ideal.
+
+For users who search RubyGems to find relevant projects, they are often presented with many options which have not been updated for many years. This presents a bad impression of the Ruby community as "unmaintained". This problem will only get worse over time if no effort is made to address the underlying reasons.
+
+We propose to introduce gem custodians and a process for approving ownership requests where the original owner is unresponsive or uncooperative. This would be an accountable process to deal with abandoned gem namespace and enable valuable gem namespace, which is effectively abandoned, to be reused in a way which is ultimately positive for the Ruby community.
```
sonalkr132, line 9: I am not sure I understand the importance of a new gem custodians group. rubygems admins should reserve the right to make the final decision.
IMO, the issue with a new group would be ensuring their motives align with rubygems. Further given that most likely it will be a volunteer position, they may not have the bandwidth of response/resolution within acceptable deadlines (which is currently offered by rubygems support).

text/0007-ownership-custodians.md
```diff
@@ -0,0 +1,113 @@
...
+
+A gem is considered "in use" if the gem itself, or any of its direct reverse dependencies (version specific) have more than 10,000 downloads in the past year.
+
+A gem is considered "maintained" if the author has logged into <https://RubyGems.org> within the past 12 months.
+
+A gem is considered to be "valid" if the homepage and other related metadata is valid (e.g. URLs return relevant pages).
+
+A gem is considered to be "working" if a user can check out and run the unit tests on a supported, non-EOL Ruby implementation.
+
+A gem is considered to be "stable" if it has a published release `>= 1.0.0`.
```
sonalkr132, line 51: These are great ideal criteria. Unfortunately, the reality is a bit nuanced. I would recommend removing following criteria:  
>  the homepage and other related metadata is valid

92% of all version have empty metadata.

>  non-EOL Ruby implementation.

more than 80% of traffic we receive is from ruby version which have reached EOL https://ecosystem.rubytogether.org/

>  it has a published release `>= 1.0.0`.

rubocop didn't have a stable release for over 8 years.

IMHO, we should stick to activity-related metrics like owner responsiveness and recent releases. Downloads can be a good measure if we filter the noise. Further as of now rubygems.org only tracks total downloads (not downloads over a period).

sonalkr132:
Npm disputes is a good reference. I also want to highlight PEP 541 https://www.python.org/dev/peps/pep-0541

Fryguy:
> EDIT: Just noticed this idea is also in #31

Not to derail this thread but

> indirect closed this 6 days ago 

😕

text/0007-ownership-custodians.md
mensfeld: @sonalkr132 we could also map out ownership and owners activities as a factor. I do it already for internal data mining. When owners are active, the change that their packages are not "abandoned" per se is higher. This coulda act as one of the signals.

text/0007-ownership-custodians.md
```diff
@@ -0,0 +1,113 @@
...
+
+For users who search RubyGems to find relevant projects, they are often presented with many options which have not been updated for many years. This presents a bad impression of the Ruby community as "unmaintained". This problem will only get worse over time if no effort is made to address the underlying reasons.
+
+We propose to introduce gem custodians and a process for approving ownership requests where the original owner is unresponsive or uncooperative. This would be an accountable process to deal with abandoned gem namespace and enable valuable gem namespace, which is effectively abandoned, to be reused in a way which is ultimately positive for the Ruby community.
+
+## Motivation
+
+There are several cases where the current gem ownership request process can be inadequate.
+
+- **Deceased Owners**: Sadly, from time to time, gem owners will pass on. In this case, it is not expected that there would be any response from the owner. We need to consider both popular and unpopular gems. Appropriate mitigating factors (like multiple authors) may not be in place when the unfortunate situation occurs.
```
svoop, line 15: Maybe repo features such as [successor settings on GitHub](https://github.com/settings/admin) might be helpful here. Given a successor is defined, the communication for transfer would be possible in the event of death without the need for a more laborious process by Rubygems which would only kick in as a fallback.

bogdanRada:
I have a little question about this RFC. I would appreciate if anyone could help me. 

Is this RFC only for public repositories on Github? I am assuming yes. Also does it apply only to gems hosted on rubygems.org?  I am assuming yes. 
   My other question is if i move everything i ever published on rubygems.org to a private gem server (possibly using geminabox) and move also all repositories somewhere else..possibly on Bit Bucket or Gitlab.. Does this RFC still apply? I am hoping it will NOT apply in this case. Or perhaps i need to host everything on a private server? 

  I don't like several things in this RFC so i am just thinking of an alternative to host my gems and repositories. Or perhaps is there a way to opt-out of this RFC? 

 Thank you very much.

simi:
@bogdanRada from my understanding this RFC applies to all gems hosted at rubygems.org. There is no connection in between GitHub and this RFC.

> I don't like several things in this RFC so i am just thinking of an alternative to host my gems and repositories. Or perhaps is there a way to opt-out of this RFC?

Can you be more specific? Which part you don't like?

nateberkopec:
Can this be closed since https://blog.rubygems.org/2022/01/19/rubygems-adoptions.html has been adopted?

sonalkr132:
I am not sure we should close this because adoptions was released. Adoptions flow solves the gem transfer/abandoned issue only if the existing owner is responsive.
We do need a policy for actions to take if the owner is unresponsive. IMHO, we should update this RFC and title to address unresponsive owners.

ioquatix:
I think a couple of things could be useful to define based on my experience.

- Owners who are unresponsive.
- Owners who are squatting = code no longer works, is maintained, or shows any sign of life, but refuse to do anything or give up the name.

svoop:
@ioquatix I'd add a third one "Owners who have passed away". While technically covered by "unresponsive", a sudden death (accident etc) doesn't give the owner the opportunity to organize an adoption in time is different enough a case to be handled in its own way: Successor settings on [GitHub](https://github.com/settings/admin) or other services which will implement similar feature in the future.

ioquatix:
@svoop it was already covered in the original proposal. :)

mullermp:
@Fryguy @ioquatix Both of you may be interested in this, given current discussions here and in other issues. https://github.com/rubygems/rfcs/pull/40

ioquatix:
@mullermp thanks. How can I help?

Issue #35, opened 2021-11-16
Require MFA for most-used gems


jchestershopify:
## Introduction

MFA is an ongoing subject of work for Rubygems, and for good reason. Research has shown[0] that account takeovers are the second-most common form of attack on software package repositories (after typosquatting). As a class of countermeasures, MFA is the best single way we have to reduce the probability of account takeovers.

We are not alone in facing this difficulty. Most notably, NPM has had a recent high-profile spate of takeover attacks on popular packages, which were only detected because of errors in the embedded exploit code. Other cases may well be undetected in the wild.

Yesterday, NPM folks announced that [they will make MFA mandatory](https://github.blog/2021-11-15-githubs-commitment-to-npm-ecosystem-security/#automated-malware-detection-and-requiring-2fa-to-combat-maintainer-account-takeovers) for "a cohort of top packages":
> ... we will begin to require two-factor authentication (2FA) during authentication for maintainers and admins of popular packages on npm, starting with a cohort of top packages in the first quarter of 2022. We are currently evaluating next steps to ensure that the strongest and most user-friendly authentication options, such as WebAuthn, are available and accessible to developers using npm.

I think Rubygems should follow this precedent. 

## Some Implementation Questions

I recognise that it's not as simple as flipping a toggle, or even doable in a single PR. But I believe we should at least aim for this policy and take incremental steps towards it.

Some problems that need to be solved for this to become a policy:

* Which gems are selected for the policy? Top _n_-thousand most-downloaded?
  * What about highly-downloaded gems for which there has been no login in a long time?
* When is the policy enforced? At login? At push?
* How do we deal with the CI/CD usecase, for which human intervention is often seen as undesirable?
* Do we need to finish WebAuthn support first, so that those using USB tokens and biometric tokens are covered?
* How is the policy announced? Blogpost? Email to owners of selected gems?
* When would the policy take effect?
* Should the policy be rolled out in phases? Eg. simple nagging at first, followed by hard enforcement later?
* Should the policy be rolled out in stages? Eg. top 10 gems, then top 100 gems, and so on?

## Next steps

The team I belong to at Shopify would like to develop an RFC for this policy change. We are  enthusiastic about raising the MFA bar for Rubygems and we'd be prepared to do a lot of the technical work in support of an accepted RFC.

[0] See [_Towards Measuring Supply Chain Attacks on Package Managers for Interpreted Languages_](https://arxiv.org/abs/2002.01139):
[...]

PR #36, opened 2022-01-12
RFC: Require MFA on the top 100 most downloaded gems

jenshenny:
Here’s the RFC that was mentioned in [this issue](https://github.com/rubygems/rfcs/issues/35) on a policy to require MFA on a cohort of gems. 

Contributors: @bettymakes, @aellispierce, @jchestershopify, @jenshenny

We’re all available on the Bundler Slack to discuss this also!


jenshenny:
Even though Shopify will be able to commit a certain number of developers to work on this policy if it's accepted, we acknowledge that there is time that needs to be spent by the Rubygems maintainers to oversee the contributions and policy changes. We would be happy to discuss ways to reduce this time and workload as much as possible.

text/0000-mfa-rollout.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Recommending MFA on Most Downloaded Gems
+- UI: https://github.com/Shopify/rubygems.org/pull/3
+- CLI Rubygems.org: https://github.com/Shopify/rubygems.org/pull/9 
+- CLI Rubygems: https://github.com/Shopify/rubygems/pull/1 
+
+## Enforcing MFA on Most Downloaded Gems
+- UI: https://github.com/Shopify/rubygems.org/pull/2 
+- CLI: https://github.com/Shopify/rubygems.org/pull/1
+
```
jenshenny, line 107: Links are still a WIP - will be updated with better prototype links to look at within the next day or two

text/0000-mfa-rollout.md
```diff
@@ -0,0 +1,129 @@
...
+When the API key is expired, the API key will be marked as invalid on the API keys page. Editing the key would be prohibited and the user can delete the key.
+
+Up for discussion, but we propose that all existing keys will be subject to a certain expiry time of around 6 months from the time of the rollout, and all new push keys are required to set an expiry date. These expirations would be staggered to avoid a thundering herd effect.
+
+### Push multiple gems at once
+
+When publishing a gem, authors can push multiple gems at once using `gem push <gems> [options]`. If the user has MFA enabled, an OTP code is to be entered once. Only one host may be specified for all gems. The response would include the results of pushing each gem (success or error). This is for gem owners that own many gems and would need to enter an OTP per gem push which can be cumbersome.
+
+## Phase 2: Recommending MFA on Most Downloaded Gems (est. Spring 2022 - Fall 2022)
```
yob, line 46: For easier reading by us southern hemisphere folks, can the timeline be specified without seasons? eg. April - October 2022.

bettymakes: That's a great point. Thanks @yob! We'll update 👍 .

simi:
I went just briefly thru the RFC and I must say it is really nice written and it makes total sense. The only problem I see is scope of `the top 100 most downloaded gems`. That is rolling scope. Gems can go in and out over the time.

Instead, I think we can make the threshold based on some metric and only add new gems to the scope. For example we can find threshold of total downloads for current top 100 gems and make it the rule. 

When total amount of downloads of gem reaches XY, we will enforce MFA to publish new release. We can also ping the authors not using MFA in advance when 90% (or any reasonable part) of threshold is reached.

mensfeld:
@simi I though about the same. Since it's a supply chain issue, if we decide to "2FA lock" a package, its dependencies should be locked as well. Otherwise we leave weak links in the chain for potential exploitation.

I am aware, that the dependency states change across the versions, so I would suggest locking the mostly used versions (if we have those metrics) based on one last year + any newly introduced.

This may also require monitoring of changes introduces to the dependency tree of "2fa locked" packages. I am willing to help with that as I have extensive experience in mining this data.

rst:
Agree that 2fa-locking dependencies of the "top 100" makes the policy a whole lot more effective.  But that puts restrictions on the maintainers of those top 100 packages which could get awkward for them.  Most notably, they'd have to be barred from adding a new dependency until its maintainer got 2fa.  It's probably worth thinking through the workflow here, for all parties, and making sure it lands in an acceptable place.

tiegz:
Love this!

Do we know how effective locking access to Rubygems will be as an enforcer here? e.g. how many packages in the top 100 aren't maintained regularly, which would prevent the maintainer from being aware of the lockout?

Any thoughts on publicly expsoing the MFA status on the gem page (or maybe the `rubygems_mfa_required` status since it's already public?)? Obviously it makes the gem more of a target and I'm not aware of another platform that does this, but it could also prompt consumers of gems to urge the maintainers to enable it.

mullermp:
Hey there, I've updated the the `awscloud` account to have MFA with the "UI and gem signin" level. Using MFA for API is a bit tricky because of CI/CD and would require larger effort. Also, keys scoped to each gem might be infeasible for us because 1) we maintain many gems, 2) we often need to access all of them at the same time, and 3) add generate, store, and retrieve new ones for each service launch.

@hsbt - I got your request from Alex Wood who no longer works on AWS SDK For Ruby. Feel free to create an issue [on the main repo](https://github.com/aws/aws-sdk-ruby) if you need us :D

jenshenny:
The CI/CD case is definitely tricky. Enforcing keys to be scoped to a gem on a [non-MFA API key](https://github.com/rubygems/rfcs/blob/e64759c550574b8c1b20ac8627459d71500ad14d/text/0007-mfa-rollout.md?plain=1#L123) might be too restrictive as you said @mullermp, and a better alternative would need to be thought out. The Shopify Rubygem account is in a similar situation. With that being said, requiring users to have the `UI and gem signin` level is a good first step for this path.

jenshenny:
> The only problem I see is scope of the top 100 most downloaded gems. That is rolling scope. Gems can go in and out over the time.
>
> Instead, I think we can make the threshold based on some metric and only add new gems to the scope. For example we can find threshold of total downloads for current top 100 gems and make it the rule.

To come back to this @simi, in the [recommended prototype](https://github.com/Shopify/rubygems.org/pull/10/files), to address the issue of gems going in and out of scope is to run a [task](https://github.com/Shopify/rubygems.org/blob/4dc97b6d184ae849167b1cdce14e58c89c6763da/script/require_mfa_for_users) to set a date for mfa enforcement for each user that fits in the metric. This task could be ran periodically to capture new gems entering the scope. According to the proposal, we selected a 6 mo period for the top 100 most downloaded gem accounts (can be changed). Would that approach work well? 

Based on your suggestion of finding a threshold of total downloads for current top 100 gems and make it the rule, I assume the alternative would be to check if it reaches this threshold when a gem download is incremented?

> When total amount of downloads of gem reaches XY, we will enforce MFA to publish new release. We can also ping the authors not using MFA in advance when 90% (or any reasonable part) of threshold is reached.

From the point of when we set the date on enforcement on a user, they will get messages on the [CLI](https://github.com/Shopify/rubygems.org/pull/9) and [UI](https://github.com/rubygems/rfcs/blob/e64759c550574b8c1b20ac8627459d71500ad14d/text/0007-mfa-rollout.md?plain=1#L46-L56) to set up MFA until the required date is reached. Do you think it's just sufficient enough to just ping the user (I'm guessing via email) instead of these prompts?

Let me know if you want me to add a description of the technical approach in the md instead of having them in the linked prototype PR descriptions.

simi:
What to do when user rejects to use MFA? Does it mean user is going to lose access to releasing gems?

What about to keep the possibility to release gem, but mark the gem as "unsafe" and let the end-user know during installation this gem was not released in secured way? That way we don't need to enforce users, but we will inform end-users, which can go and ask maintainer to release in secure way. Clearly there will be warning during gem push (we can also make it fail in later phase without explicitly passing `--insecure` flag to gem push for example).

_By my experience, enforcing MFA (or actually enforcing anything in general) is painful process for users. Making it optional, but little more "painful" when not used seems like better option to me._

jenshenny:
> What to do when user rejects to use MFA? Does it mean user is going to lose access to releasing gems?

In the current state, yes. You raise a good point that users can’t release gems especially in an emergency situation.

At this moment, I’m open to the possibility of marking unsafe on gems. This is something we can do for the entire rubygems population vs a cohort (it would be confusing if 1. a gem without mfa in a subset is marked unsafe, vs 2. others that aren’t in the subset, doesn’t have mfa but isn’t marked unsafe). 

However, from this [discussion](https://github.com/rubygems/rubygems.org/discussions/2889), I’m not sure if this will encourage shaming if we want to inform end users during installation especially for a highly downloaded gem, though I guess it might be warranted if prior notice was given. It could do more harm than good and prolong adoption. I’m leaning towards using this solution for the general population and catch gem dependencies to enable MFA, but still require MFA on the most frequently used gem user accounts if adoption isn’t high enough before marking things unsafe.

jchestershopify:
I agree with @jenshenny -- I don't think a warning and shaming will significantly move the needle.

What I'd like to do is reframe how we think about security issues like MFA. I think end users should be at the centre of every decision and their needs should be put first.

For example, consider an account takeover. For a gem maintainer, an account takeover is distressing and embarrassing. But for end users, it is potentially catastrophic. And end users outnumber maintainers by a large ratio. In terms of a calculus of harm, I think end users of gems deserve to be given priority. That means that they need maintainer accounts to be secured with MFA.

text/0000-mfa-rollout.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+## Phase 4: Rolling out to the General Public (2023)
+
+Once the rollout of the top 100 gems is successfully rolled out, a similar rollout will be done to another cohort of gems until all accounts are required to have MFA enabled. 
+
+# Reference-level explanation
+
+## API Key and Gem Publishing Improvements
+The API key model will need to be changed in order to implement these improvements. For new features like adding gem scope and expiry date on gems, columns will be added for expiry and gem scope and additional options will need to be added to the CLI and the UI. Setting an expiry on existing keys can be considered a breaking change.
```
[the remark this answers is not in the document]
jenshenny: Thanks @tiegz !

jenshenny:
NPM announced that they are requiring MFA to the [top 100 npm package maintainers](https://github.blog/2022-02-01-top-100-npm-package-maintainers-require-2fa-additional-security/ ) that started last week, and rolling it in phases to all accounts by March. I believe Rubygems should also follow NPM's lead and follow a similar policy and rollout. If a maintainer doesn't want to enable MFA, we can make it a bit more annoying by making them enter an OTP sent to their email on the UI and `gem signin`.

schneems:
## My experience

I've been using MFA on all my gems https://rubygems.org/profiles/schneems for several years now. My employer, Heroku, required it. I think it's a nice layer of protection and while it does add another step to releasing, it's (to me) a small one.

I release via `rake release` with the bundler release tasks. It prompts me for a MFA code. I've got that set up on my laptop already so I open an app, put in a passcode, type in "ruby" copy and paste the MFA in and then I'm done. I could probably automate it more to require fewer steps, but I've not (personally) felt the need.

## My opinion

It's my opinion/experience that MFA is a large bang-for-your-buck and is already expected in many/most places.

I would like us to be "secure by default" as much as possible. 

There are some implementation questions around how to get there. If someone "rejects" the requirement to use MFA they can run their own RubyGems server to host their own gems and release in that way. If a gem says "add this other source to your Gemfile" that's at least a traceable artifact where someone could visually audit the Gemfile and identify all secondary sources that might not be as secure.

I would support a rolling process for making all gems MFA as it only takes one broken link to have a supply chain attack. Starting at the first 100 most popular is a good step, but doesn't go far enough (again, my opinion). If there's an escape valve of "run your own gemserver if you really hate it that much" I think that handles edge cases. I do think it's a good idea to think about the edge cases, but as long as there is at least one escape valve that covers them, I don't know that we need to overly go out of our way to try to please everyone.

I think with security stuff it's a balance between things that reasonably reduce attack surface area versus usability. One example of this going wrong might be a company that aggressively enforces password rotation frequently and then employees start writing their password down by their desk because they don't have enough time to memorize it. It's important to think on possible unintended consequences, but we also will not come to a case of 100% consensus of all cases perfectly handled.

text/0007-mfa-rollout.md
```diff
@@ -0,0 +1,125 @@
...
+
+Without requiring MFA on user accounts, there is a greater risk of malicious gem releases due to account takeovers on the Rubygems platform. This will cause a larger burden to the owners of the malicious package version, end users and Rubygems maintainers when these attacks occur.
+
+# Unresolved questions
+
+- Legacy keys are a relic of pre Rubygems 3.2 but still are in existence. What would the impact be of removing legacy key support (it comes with removing Rubygems <3.2 also)?  Legacy keys bring a larger risk to accounts and can bypass the restrictions we are placing on API keys. More specifically, how many legacy keys are still being used recently among the 100 most downloaded gem accounts and the greater ecosystem? Are there any metrics determining the amount of people in Rubygems <3.2?
```
sonalkr132, line 122: Legacy keys are not in use, at least from the perspective of rubygems.org. All legacy keys were migrated to the new API  key and will support any limitation added on MFA. The GET endpoint used in clients older than 3.2 also creates the new API key.

sonalkr132:
>  I assume the alternative would be to check if it reaches this threshold when a gem download is incremented?

I think using the download threshold would be more straightforward. https://rubygems.org/gems/sinatra is the 100th gem, we can use its count as the threshold.

Looking at top 100 gems pusher (of latest_version) MFA status, I can see that 72 already have MFA level in ui_and_api and ui_and_gem_signin. 11 don't have pusher, because they haven't had a release since started we storing pusher https://github.com/rubygems/rubygems.org/pull/1954/. This means proposed changes here only affect 17 users. I can share the list privately to someone here if they want to reach out to these users privately to check their use-case (through twitter?) .
Given that we already have good adoption, I think we should shrink the timeline. Start showing the warning to these users ASAP and enforce by july-aug 2022.
I would also add writing a blog post explaining the changes and perhaps adding a job to send email reminders of coming changes to affected users.

sonalkr132:
I also want to highlight this post https://thehackerblog.com/zero-days-without-incident-compromising-angular-via-expired-npm-publisher-email-domains-7kZplW4x/ 
We have checks in place to mitigate it however it is not full proof, MFA is the ultimate solution.

hsbt:
I'm targeted maintainers for this proposal. I understood the security is an important part of language ecosystem. So, I enforced to use MFA authentication in https://github.com/ruby and https://bugs.ruby-lang.org/.

I'm frustrated to enforce something for my OSS work from a big tech company's order. But I agreed to this proposal.

text/0007-mfa-rollout.md
```diff
@@ -0,0 +1,150 @@
...
+
+Gem publishing will be improved by allowing gem owners to push mulitple gems at once, so that they only need to enter their OTP code one time. A prototype for pushing multiple gems has been briefly explored here:  https://github.com/Shopify/rubygems/pull/2
+
+## Policy Rollout
+A user account will be required to have MFA enabled if they own a gem that exceeds a download threshold of 175 million. This number was selected as it is around the total downloads of the 100th most downloaded gem ([sinatra](https://rubygems.org/gems/sinatra)). There will be two methods introduced in the Rubygems model. One would determine whether the gem owners are encouraged to setup MFA (eg. `mfa_recommended?`) and display relevant warnings on the UI/CLI. The other would determine if gem owners are required to setup MFA (eg. `mfa_required?`) and start blocking sensitive behaviours.
```
jenshenny, line 99: > We can also ping the authors not using MFA in advance when 90% (or any reasonable part) of threshold is reached. (from https://github.com/rubygems/rfcs/pull/36#issuecomment-1024894144)

I don't have much context around the rate of downloads but would it make sense to warn/ping users ASAP at around ~175 million, and enforce at ~180 million in July/Aug (which hopefully encompasses `sinatra` at that point). This would prevent owners on the border to be surprised once they reach the required threshold.

jenshenny:
@simi @sonalkr132 Bumping this up to see if you have any feedback on the above changes.

simi:
> Bumping this up to see if you have any feedback on the above changes.

Thanks for the ping @jenshenny. 

> I'm frustrated to enforce something for my OSS work from a big tech company's order. But I agreed to this proposal.

I do hear you @hsbt. I'm thinking how to prevent this kind of feeling for other gem owners as well. If I understand it well, nothing is being enforced for now, right? We need to ensure that "enforcing" part would be smooth as possible to not make anyone feeling this way.

I think we basically all agree on phase 1 and 2. I was thinking how to move this forward. What about to split this RFC into two parts? Preparation + recommended part (phase 1 and 2) and "enforcing" part (phase 3 and 4). That way I think we can consider first part as agreed and merge this part of RFC. We can still do some small tweaks in there, but the overall goal would stay the same.

And for the second part I suggest to open separated RFC (we can scope this one only to first part) and we can continue in there. I think we can keep the dates open unless we find smooth soultion for enforcing.

WDYT @jenshenny @hsbt @sonalkr132 @jchestershopify @tiegz @schneems @mensfeld @rst?

mensfeld:
@simi I agree with the enforcement but I also see @hsbt point of view though I disagree with it. I think that 2fa in the case of popular packages should just become a standard.

I think improvements to the gem publishing flow as well as recommendations should be implemented as soon as doable for the benefit of the ecosystem. So phases 1 and 2 are :+1:  from me for sure.

@sonalkr132 
> Anything below 20k downloads would fall this category for me (includes 83% of gems).

Then we have a case of new dependencies coming to the supply chain while not being popular (yet). For example, when `faraday-rack` was published not long ago (last year) it immediately became transitive dependency for many projects while still being (for a short period of time) < 20k. Projects are merged and splitted all the time and download count may not be the best metric for new packages being introduced into the chain.

Maybe we should make it not based on downloads but rather based on a collective dependency tree popularity? That way we would protect the whole supply chain for popular components (collectively) while not introducing any burden to private projects.

jchestershopify:
> From the discussion here however, I imagine using a download threshold is easier than using a dependency tree.

That's my belief also. They'll be fairly well correlated in the long run, even if in the short run (such as `faraday-rack`) there could be outliers.

mensfeld:
@jchestershopify It's not often that the chain is long. Calculating 1 or 2 previous links for the "forcing 2fa" would be trivial for new packages. For older downloads count would be enough. Even with the current setup of rubygems it would require at most 2 additional calls to the db (I may be wrong on it though) for this data and only in case of new versions of packages that are below threshold.

sonalkr132:
In my experience, dependency chain is not as easy to calculate. It would inevitably need recursion and we won't be able to add it for "on the fly" check. You also need to consider (all?) versions, for example:
- A is one of top 100 gem
- 1.0 of A requires ~> "1.0" of B
- 1.1 (latest) of A does not require B
- B has active development on 0.9.x and 1.0.x series

Do we inforce mfa requirement on B even if latest of A doesn't require B? Do we inforce mfa on 0.9.x release of B as well even if it is not used by A?
If we add an offline process to mark gems for mfa requirement, how do we ensure that we don't have flapping issue simi brought up.

text/0007-mfa-rollout.md
bettymakes: I like the idea of introducing a signal earlier before gem owners have reached the 175 million threshold. If we do implement this, I'd recommend warning them sooner than ~170 million downloads since by that point they'll be ~97% away from reaching the ~175 million threshold. 

The guidance of ~90% away from threshold seems reasonable. That would translate to warning ~157 million downloads.

jenshenny: 👍 That makes sense, I updated the thresholds to 180 million for required since `sinatra` (100th most downloaded gem) has reached that mark and 165 million for recommended. 165 million is roughly 90% of 180 million.

jenshenny:
@simi you suggested to only merge phases 1 and 2. How do you feel about merging the RFC in its current state with part 3 included since there’s support in doing so in subsequent comments? 

Are there any other blockers that are noticeable regarding merging the RFC in this current state (with phase 4 discussed in another RFC)? 

> Perhaps we should check if we were enfore this only top 100 and all their transitive runtime dependencies, how many gems/users will it affect?

Using [this script](https://github.com/Shopify/rubygems.org/blob/02fb7001741b02bf5d4e473750a3edca6b653ad9/script/transitive_deps) (which I hope is correct) with today's data dump, the total number of dependencies it'll affect would be ~750. Only counting runtime dependencies, there’s 425 and the least downloaded gem is [aws-keyspaces](https://rubygems.org/gems/aws-sdk-keyspaces) which was created a week ago. The one after has more than 850 000 downloads. If we continue this rollout on a rolling basis and stop at the download threshold for personal projects, I think it would work well.

simi (approved):

simi:
@jenshenny thanks for moving this forward. RFC looks good to me now. Since phase 4 would be specified later in separate RFC and for phase 3 there is just estimate date (and also since nothing is set into stone even after we merge this) I'm ok to merge this.

To follow the process (https://github.com/rubygems/rfcs#what-the-process-is), should we call for `final comment period`? @indirect @sonalkr132 @deivid-rodriguez any idea?

simi:
OK, let's do it.

simi:
:partying_face:

simi:
@ioquatix is there anything we can do to help release Unicorn? Is there any info why current MFA options are not usable for given usecase? I think it would be better to improve MFA and make it easier to adopt instead of adding additional feature to bypass the requirements.

indirect:
One easier to use type of MFA that we have not (yet?) implemented is: send a code in an email, request the code to be entered. Since RubyGems.org accounts already require an email and password, it seems like that should be usable by anyone who already has a RubyGems.org account. @ioquatix do you know if that would suffice?

ioquatix:
We'd have to contact Eric to see what would work for him.

PR #40, opened 2022-04-05
Scoped gems proposal

mullermp:
[rendered proposal](https://github.com/mullermp/rfcs/blob/master/text/0008-scoped-gems.md#rubygems-scope)

Hello RubyGems team, and all those that come across this.

In this PR, I have included a proposal for a feature called "scoped gems". In short, the proposal is to widen the gem naming specification to include a new character `@` to group related gems together under a specific organization reserved suffix. The naming pattern follows `gem_name@scope`. On the first gem push (new record), if the gem is scoped (follows the pattern), the gem's scope will be validated to have been created by a user from an organization that reserved the scope. A scoped gem can be installed and required as normal gems are today.

For example, consider `aws-sdk-s3`, the S3 gem for AWS. If this gem were scoped, it could be published as `s3@aws-sdk` (or more generically, `<service>@aws-sdk`). This gem can only be created by a user in the AWS organization (which has reserved the `aws-sdk` scope). A user can install this gem with `gem install s3@aws-sdk` and require it with `require 's3@aws-sdk'`.

The main benefits of this feature are that organizations can publish their own groups of gems (i.e. multiple organizations can have a "configuration" gem), and organizations are able to reserve gem names (via `@scope` suffix, similar to a reserved prefix). A developer can be reasonably sure that any new gem such as `new-cool-feature@rails` is an official Rails gem, or `new-s3-service@aws-sdk` is an official AWS SDK client, or even `socket@ruby` to be an official stdlib gem! This reservation system combats "fake" "similarly named" gems that are branded as official that attempt to steal personal information.

Please leave any feedback and I would be happy to amend the approach/design.

### DIFF
--- a/text/0008-scoped-gems.md
+++ b/text/0008-scoped-gems.md
@@ -1,0 +1,167 @@
+# RubyGems Scope
+
+* Feature Name: Scoped Gems
+* Start Date: 4/1/2022
+* RFC PR:
+* Bundler Issue:
+
+# Summary
+
+Scoped gems is a feature that allows for grouping related gems together under a specific user or organization account. Scoped gems are identified via their name following a specific pattern, and can be built, installed, and required the same way as existing gems today. Each user account has their own scope, and only that user account is permitted to register a new gem under that scope (first-time publish); Once one version is published, any owner of the gem may publish new versions.
+
+# Motivation
+
+There are three main motivations for supporting scoped gems:
+
+1. Widen the availability of gem names. Many users may want to publish a Configuration gem, but the name configuration has already been reserved. With scoped gems, each user can have their own configuration gem in their own scope.
+2. Reduce customer confusion for tools and services spanning multiple gems. It is not uncommon for an organization to distribute many packages, such as Rails with the ActiveX packages, AWS with SDK clients, etc. Scoped gems are a good way to signal official packages for organizations.
+3. Help reduce [typo-squatting](https://blog.reversinglabs.com/blog/mining-for-malicious-ruby-gems) with malicious code. Scoped gems are also susceptible to typos, however, gems published under a scope are verified to be owned by that user.
+
+# Guide-level explanation
+
+A scoped gem is any gem that is named following the pattern `gem@scope`, where “gem” is the name of the gem, and “scope” is the user’s scope (username or organization account).
+
+As an example, consider a scoped gem called `hola@mullermp.gemspec`. Note that both the gemspec’s `name` and the required file `lib/hola@mullermp.rb` follows the `gem@scope` pattern:
+
+```
+Gem::Specification.new do |s|
+  s.name = 'hola@mullermp'
+  s.version = '0.0.0'
+  s.summary = "Hola!"
+  s.authors = ["Matt Muller"]
+  s.files = ["lib/hola@mullermp.rb"]
+end
+```
+
+The scoped gem can be built with `gem build`:
+
+```
+$ gem build hola@mullermp.gemspec 
+  Successfully built RubyGem
+  Name: hola@mullermp
+  Version: 0.0.0
+  File: hola@mullermp-0.0.0.gem
+
+```
+
+And the scoped gem can be installed via command line:
+
+```
+$ gem install hola@mullermp-0.0.0.gem
+Successfully installed hola@mullermp-0.0.0
+Parsing documentation for hola@mullermp-0.0.0
+Done installing documentation for hola@mullermp after 0 seconds
+1 gem installed
+```
+
+or installed via Bundler:
+
+```
+$ cat Gemfile
+gem 'hola@mullermp'
+
+$ bundle install
+Resolving dependencies...
+Using bundler 2.2.22
+Using hola@mullermp 0.0.0
+Bundle complete! 1 Gemfile dependency, 2 gems now installed.
+Use `bundle info [gemname]` to see where a bundled gem is installed.
+```
+
+When installed, developers can expect that the gem `hola` comes from a known source user or organization, `mullermp`, because the [RubyGems.org](http://rubygems.org/) website validated the first-time publish of `hola@mullermp` to be from `mullermp`.
+
+Finally, in Ruby code, the gem can be required:
+
+```
+$ irb
+irb(main):001:0> require 'hola@mullermp'
+=> true
+irb(main):002:0> Hola.hi
+Hello world from my Scoped gem!
+=> nil
+```
+
+# Reference-level explanation
+
+There are two types of changes required, one change to the `gem` command tooling and another on the [RubyGems.org](http://rubygems.org/) API.
+
+## `gem` commands
+
+To support the pattern in the `gem` commands, the `VALID_NAME_PATTERN` regex should be updated to include the `@` character.
+
+```
+VALID_NAME_PATTERN = /\A[a-zA-Z0-9\@\.\-\_]+\z/.freeze # :nodoc:
+```
+
+This change allows for `hola@mullermp.gemspec` to be built, and for `hola@mullermp-0.0.0.gem` to be installed.
+
+## [RubyGems.org](http://rubygems.org/) API
+
+Similar to the `gem` commands, the `NAME_PATTERN` regex in [patterns.rb](https://github.com/rubygems/rubygems.org/blob/84a49869c19303bf3a02c9f7edab67de093c65c1/lib/patterns.rb) needs to be updated:
+
+```
+# Adds @ for scope support
+ALLOWED_CHARACTERS = "[A-Za-z0-9**@**#{Regexp.escape(SPECIAL_CHARACTERS)}]+".freeze
+```
+
+In the [Rubygem model](https://github.com/rubygems/rubygems.org/blob/master/app/models/rubygem.rb), we can make use of the `needs_name_validation?` predicate:
+
+```
+validate :ensure_gem_scope, if: :needs_name_validation?
+```
+
+and `ensure_gem_scope` might look something like:
+
+```
+# Only invoked when needs_name_validation? is true
+# Which is on new_record? or name_changed?
+def ensure_gem_scope
+  gem, scope = name.split("@")
+  return unless scope
+  if scope != user
+    # Raise an error and/or prevent gem push
+  end
+end
+```
+
+# Drawbacks
+
+One main drawback is confusion between normal gems and scoped gems. It is important that we update the [recommendations](https://guides.rubygems.org/name-your-gem/) to explain the scoped gems feature and what Ruby namespaces to expect.
+
+An obvious drawback is, with widened gem name availability, there may be more namespace conflicts. Consider a case where user1 creates `configuration@user1` and user2 creates `configuration@user2`, both creating a `Configuration` class. If user 3 installs and requires both, then there may be side effects. This behavior exists today albeit gem names may be different.
+
+# Rationale and Alternatives
+
+The main benefits of this approach are:
+
+1. It is a frugal change to the RubyGems gem specification that is mostly a naming standard. It removes validation around gem naming on `gem build` and `gem install`, and adds validation for first-time gem publishing via `gem push`.
+2. Avoids changes to the Ruby library (and its versions) and how gems are required.
+3. Similar to reserved prefixes for verified publishers (like [NuGet](https://github.com/NuGet/Home/wiki/NuGet-Package-Identity-Verification)) but it can be considered as a *reserved suffix*.
+4. Possible removal of “blacklisted” gem names as important gem names can be migrated to scopes, i.e. `socket@ruby`.
+
+## Alternative 1 - Reserved Prefixes
+
+An alternative to this approach would be to reserve prefixes, like `aws-sdk-` for AWS SDKs. The major upside is that there’s no change to the gem specification and can be implemented even more cheaply. However, the downsides to this approach are:
+
+1. Prefix registration may involve an application and/or manual review of some kind.
+2. Gem name availability is not widened (i.e. configuration gem example), unless you reserve a prefix for your own username.
+3. Some organizations do not follow a prefixing pattern (i.e. Rails with Active/ActionX, some AWS products, etc).
+4. Enforcing new validation on an existing set of gems may result in conflicts and exceptions that need to be codified. An “open” prefixing reservation system (i.e. each user has their own prefix reserved by default) is certainly a backwards incompatible change.
+
+## Alternative 2 - Scoping with dots (.)
+
+In the current specification (although not documented), dots (`.`) are allowed. There are [a few “hidden” gems](https://rubygems.org/search?query=.) that use dots. If adding the `@` character is not desired, the dot character can be re-used to handle scoping, perhaps with the pattern `scope.gem_name` such as `aws-sdk.s3`.
+
+# Unresolved questions
+
+**How do we migrate existing gems to scoped naming?**
+Take for example `aws-sdk-s3`. A new scoped gem name might be `s3@aws-sdk`. `aws-sdk-s3` may need to have an “alias” pointing to `s3@aws-sdk` and be resolved when performing `gem install`, or `s3@aws-sdk` will need to be a new gem (duplicate publish) with an eventual deprecation of `aws-sdk-s3`.
+
+**What about old versions of the `gem` command tooling?**
+Some users will be using an older rubygems version and attempt to push or install a scoped gem. What user experience do we need or want?
+
+# Related Links
+
+Namespace gem discussion - https://github.com/rubygems/rfcs/issues/31
+Verified users proposal - https://github.com/rubygems/rubygems.org/pull/2698
+
--- a/text/0008-scoped-gems.md
+++ b/text/0008-scoped-gems.md
@@ -1,4 +1,4 @@
-# RubyGems Scope
+# RubyGems Scoped Gems and Organizations
 
 * Feature Name: Scoped Gems
 * Start Date: 4/1/2022
@@ -7,21 +7,23 @@
 
 # Summary
 
-Scoped gems is a feature that allows for grouping related gems together under a specific user or organization account. Scoped gems are identified via their name following a specific pattern, and can be built, installed, and required the same way as existing gems today. Each user account has their own scope, and only that user account is permitted to register a new gem under that scope (first-time publish); Once one version is published, any owner of the gem may publish new versions.
+ Scoped gems is a feature that allows for grouping related gems together under an organization account. Scoped gems are identified by their name following a specific pattern of `gem@scope`. They can be built, installed, and required the same way as existing gems today. An Organization is a group of users that can own 0 or many gem scopes and 0 or many gems. Only the users that are part of an organization are permitted to publish a new gem under a scope that is registered to that organization (first-time publish).
 
 # Motivation
 
-There are three main motivations for supporting scoped gems:
+There are three main motivations for supporting scoped gems and organizations:
 
-1. Widen the availability of gem names. Many users may want to publish a Configuration gem, but the name configuration has already been reserved. With scoped gems, each user can have their own configuration gem in their own scope.
-2. Reduce customer confusion for tools and services spanning multiple gems. It is not uncommon for an organization to distribute many packages, such as Rails with the ActiveX packages, AWS with SDK clients, etc. Scoped gems are a good way to signal official packages for organizations.
-3. Help reduce [typo-squatting](https://blog.reversinglabs.com/blog/mining-for-malicious-ruby-gems) with malicious code. Scoped gems are also susceptible to typos, however, gems published under a scope are verified to be owned by that user.
+1. Reduce customer confusion for tools and services spanning multiple gems. It is not uncommon for an organization to distribute many packages, such as Rails with the ActiveX packages, AWS with SDK clients, Ruby standard library gems, Azure SDKs, etc. Scoped gems are a good way to signal official packages from organizations.
+2. Help reduce [typo-squatting](https://blog.reversinglabs.com/blog/mining-for-malicious-ruby-gems) with malicious code. Scoped gems are also susceptible to typos, however, gems published under a scope are verified to be owned by that organization.
+3. An organization allows for a better gem owner permissions model. Gems owned by the same company or organization currently need to manage users for each gem. Users and gem owners can now be part of a single entity, and that entity can be the owner of the gem.
 
 # Guide-level explanation
 
-A scoped gem is any gem that is named following the pattern `gem@scope`, where “gem” is the name of the gem, and “scope” is the user’s scope (username or organization account).
+## Scoped gems
 
-As an example, consider a scoped gem called `hola@mullermp.gemspec`. Note that both the gemspec’s `name` and the required file `lib/hola@mullermp.rb` follows the `gem@scope` pattern:
+A scoped gem is any gem that is named following the pattern `gem@scope`, where “gem” is the name of the gem, and “scope” is an organization's reserved gem scope. Gem scopes are globally unique across all organizations.
+
+As an example, consider a scoped gem defined as `hola@mullermp.gemspec`. Note that both the gemspec’s `name` and the required file `lib/hola@mullermp.rb` follows the `gem@scope` pattern:
 
 ```
 Gem::Specification.new do |s|
@@ -36,7 +38,7 @@ end
 The scoped gem can be built with `gem build`:
 
 ```
-$ gem build hola@mullermp.gemspec 
+$ gem build hola@mullermp.gemspec
   Successfully built RubyGem
   Name: hola@mullermp
   Version: 0.0.0
@@ -68,7 +70,7 @@ Bundle complete! 1 Gemfile dependency, 2 gems now installed.
 Use `bundle info [gemname]` to see where a bundled gem is installed.
 ```
 
-When installed, developers can expect that the gem `hola` comes from a known source user or organization, `mullermp`, because the [RubyGems.org](http://rubygems.org/) website validated the first-time publish of `hola@mullermp` to be from `mullermp`.
+When installed, developers can expect that the gem `hola` comes from a known organization that owns the `mullermp` gem scope, because the [RubyGems.org](http://rubygems.org/) website validated the first-time publish of `hola@mullermp` to be from a user in that organization.
 
 Finally, in Ruby code, the gem can be required:
 
@@ -81,13 +83,29 @@ Hello world from my Scoped gem!
 => nil
 ```
 
+The user can *reasonably* assume that the `hola@mullermp` gem has a namespace of `Hola`, following the [recommended naming guidelines](https://guides.rubygems.org/name-your-gem/).
+
+## Organizations
+
+An organization is a group of users that can own 0 or many gems and 0 or many gem scopes. A user can visit the `mullermp` organization by visiting the following link: `https://rubygems.org/organizations/mullermp` where `mullermp` can be any organization name. On this page, a user would see a "profile" that lists the organization's members (a user named `mullermp`), the gems owned by the organization (one gem, `hola@mullermp`), and any reserved gem scopes (`mullermp` scope). If the user is part of the organization, they can register additional gem scopes, with limitations.
+
+The organization's members are effectively the the gem owners and maintainers. On a gem's page, such as `https://rubygems.org/gems/hola@mullermp`, a new section labeled as "Organization" is displayed, with a link to the `mullermp` organization page. The organization's members, are "splat" onto the gem's page like normal gem owners.
+
 # Reference-level explanation
 
-There are two types of changes required, one change to the `gem` command tooling and another on the [RubyGems.org](http://rubygems.org/) API.
+## [RubyGems.org](http://rubygems.org/) Organization MVC
+
+A standard Rails scaffold is needed for an organization entity.
+
+1) Controller - An organizations controller is needed for administration, such as: listing/adding/removing owners, listing gems owned by the organization, and reserving gem scopes.
+
+2) Model - The Organization model will need the following fields/references: a list of owners that can perform administration (user references), a list of reserved scopes that the organization owns (string list or references), and a list of gems owned by the organization (rubygem references).
+
+3) View - The Organization view will be routed under `/organizations/<org name>`. The page will have forms for adding/removing of owners and reserving gem scopes, as well as displaying any textual information such as the organization's mission statement, a profile picture, etc.
 
 ## `gem` commands
 
-To support the pattern in the `gem` commands, the `VALID_NAME_PATTERN` regex should be updated to include the `@` character.
+To support the pattern in the `gem` commands, the [`VALID_NAME_PATTERN` regex](https://github.com/rubygems/rubygems/blob/504272e35d35a5c7dcdccfece574f5d8a060b148/lib/rubygems/specification.rb#L109) should be updated to include the `@` character (in both the specification and the policy).
 
 ```
 VALID_NAME_PATTERN = /\A[a-zA-Z0-9\@\.\-\_]+\z/.freeze # :nodoc:
@@ -104,7 +122,7 @@ Similar to the `gem` commands, the `NAME_PATTERN` regex in [patterns.rb](https:/
 ALLOWED_CHARACTERS = "[A-Za-z0-9**@**#{Regexp.escape(SPECIAL_CHARACTERS)}]+".freeze
 ```
 
-In the [Rubygem model](https://github.com/rubygems/rubygems.org/blob/master/app/models/rubygem.rb), we can make use of the `needs_name_validation?` predicate:
+In the [Rubygem model](https://github.com/rubygems/rubygems.org/blob/1cb71cb167076d08fb3a81209a72af04d37da16f/app/models/rubygem.rb), we can make use of the `needs_name_validation?` predicate:
 
 ```
 validate :ensure_gem_scope, if: :needs_name_validation?
@@ -118,17 +136,49 @@ and `ensure_gem_scope` might look something like:
 def ensure_gem_scope
   gem, scope = name.split("@")
   return unless scope
-  if scope != user
+  # find the owning organization
+  org = Organization.find_by_scope(scope)
+
+  unless org.owners.include?(current_user)
     # Raise an error and/or prevent gem push
   end
 end
 ```
 
+## Migrating existing gems to an organization (scoped gems)
+
+Migration from an existing gem to a scoped gem cannot be done by RubyGems.org alone. The preferred migration strategy will be to convert un-scoped gems into a shim that forwards to a scoped gem using a gemspec dependency and a require statement.
+
+As an example, consider the un-scoped gem `hola`. This gem needs to be transferred to the `mullermp` namespace. The steps are as follows:
+
+1. Build and publish any version of `hola@mullermp` on RubyGems.org just like any other gem.
+
+2. Modify `hola.gemspec` to have one single dependency on `hola@mullermp`:
+  ```
+  s.add_runtime_dependency 'hola@mullermp' # open ended, but can use major version constraint
+  ```
+
+3. Modify `lib/hola.rb` to have a single require statement: `require 'hola@mullermp'`. Optionally modify the gemspec's `post_install_message` to warn the user about forwarding.
+
+4. Build and publish `hola`. No new versions to `hola` should be published.
+
+Legacy applications that install `hola` will also download and install `hola@mullermp`. In code, `hola` can still be required and used as normal. Attempting to require `hola@mullermp` will return false as it is already loaded.
+
+```
+irb(main):001:0> require 'hola'
+=> true
+irb(main):002:0> Hola.hi
+Hello world!
+=> nil
+irb(main):003:0> require 'hola@mullermp'
+=> false
+```
+
 # Drawbacks
 
-One main drawback is confusion between normal gems and scoped gems. It is important that we update the [recommendations](https://guides.rubygems.org/name-your-gem/) to explain the scoped gems feature and what Ruby namespaces to expect.
+One main drawback is confusion between normal gems and scoped gems. It is important that we update the [recommendations](https://guides.rubygems.org/name-your-gem/) to explain the scoped gems feature and what Ruby namespaces to expect. RubyGems.org does not enforce Ruby namespace with gem naming conventions, nor should it. The `hola@mullermp` gem is expected to have a `Hola` namespace, just as `net-http-digest_auth@ruby` (a potential Ruby official gem) would have `Net::HTTP::DigestAuth`.
 
-An obvious drawback is, with widened gem name availability, there may be more namespace conflicts. Consider a case where user1 creates `configuration@user1` and user2 creates `configuration@user2`, both creating a `Configuration` class. If user 3 installs and requires both, then there may be side effects. This behavior exists today albeit gem names may be different.
+An obvious drawback is, with widened gem name availability, there may be more namespace conflicts. Consider a case where org1 creates `configuration@org1` and org2 creates `configuration@org2`, both creating a `Configuration` class. If a user installs and requires both, then there may be side effects. This behavior exists today albeit gem names may be different. Developers will need to continue using discretion.
 
 # Rationale and Alternatives
 
@@ -136,8 +186,8 @@ The main benefits of this approach are:
 
 1. It is a frugal change to the RubyGems gem specification that is mostly a naming standard. It removes validation around gem naming on `gem build` and `gem install`, and adds validation for first-time gem publishing via `gem push`.
 2. Avoids changes to the Ruby library (and its versions) and how gems are required.
-3. Similar to reserved prefixes for verified publishers (like [NuGet](https://github.com/NuGet/Home/wiki/NuGet-Package-Identity-Verification)) but it can be considered as a *reserved suffix*.
-4. Possible removal of “blacklisted” gem names as important gem names can be migrated to scopes, i.e. `socket@ruby`.
+3. Similar to reserved prefixes for verified publishers (like [NuGet](https://github.com/NuGet/Home/wiki/NuGet-Package-Identity-Verification)). This approach can be considered as a *reserved suffix*.
+4. Eventual removal of "blacklisted" gem list as important standard Ruby gem names are migrated to scopes (i.e. `socket@ruby`, no need to reserve `socket`).
 
 ## Alternative 1 - Reserved Prefixes
 
@@ -152,16 +202,25 @@ An alternative to this approach would be to reserve prefixes, like `aws-sdk-` fo
 
 In the current specification (although not documented), dots (`.`) are allowed. There are [a few “hidden” gems](https://rubygems.org/search?query=.) that use dots. If adding the `@` character is not desired, the dot character can be re-used to handle scoping, perhaps with the pattern `scope.gem_name` such as `aws-sdk.s3`.
 
+## Alternative 3 - `@scope/gem_name` pattern.
+
+While it might be preferable to use this pattern, it has implementation and legacy complications, specifically with forward slash `/`. Gem install and build will assume `@scope` as a directory and will fail. Gems installed in a location is currently flat and would have impact on existing tooling if installed within folders. Unless the `/` is escaped, the RubyGems.org API will also have issues resolving and displaying scoped gems because of Rails resource routing. This approach can be used with a more thought out design and a willingness for making large and downstream changes.
+
 # Unresolved questions
 
-**How do we migrate existing gems to scoped naming?**
-Take for example `aws-sdk-s3`. A new scoped gem name might be `s3@aws-sdk`. `aws-sdk-s3` may need to have an “alias” pointing to `s3@aws-sdk` and be resolved when performing `gem install`, or `s3@aws-sdk` will need to be a new gem (duplicate publish) with an eventual deprecation of `aws-sdk-s3`.
+**Adding `scope` to the gemspec**
+As an alternative to overloading the gemspec's name, we can consider having a `scope` attribute, i.e.
+```
+  s.name = 'hola'
+  s.scope = 'mullermp'
+  s.version = '0.0.0'
+```
+This gem would still build as `hola@mullermp-0.0.0.gem` and have the same usage experience. However, would the gemspec file be named `hola@mullermp.gemspec` or `hola.gemspec`, and what implication would that have? If using `hola.gemspec`, is there a possibility someone may work with 2 `hola` gems from different organizations (x-organizational work). If the spec's name is `hola@mullermp`, we may as well overload the spec name too.
 
-**What about old versions of the `gem` command tooling?**
-Some users will be using an older rubygems version and attempt to push or install a scoped gem. What user experience do we need or want?
+**Requirements for having an organization or for reserving gem scopes**
+To prevent a rush on reservations, organizations and/or gem scopes should be limited by some initial process. Eventually this limitation would be removed. How should this be handled?
 
 # Related Links
 
 Namespace gem discussion - https://github.com/rubygems/rfcs/issues/31
 Verified users proposal - https://github.com/rubygems/rubygems.org/pull/2698
-
--- a/text/0008-scoped-gems.md
+++ b/text/0008-scoped-gems.md
@@ -220,6 +220,9 @@ This gem would still build as `hola@mullermp-0.0.0.gem` and have the same usage 
 **Requirements for having an organization or for reserving gem scopes**
 To prevent a rush on reservations, organizations and/or gem scopes should be limited by some initial process. Eventually this limitation would be removed. How should this be handled?
 
+**Updated Regex to validate one single `@` character in the middle**
+Optionally, the `VALID_NAME_PATTERN` regex can be updated to `\A[a-zA-Z][a-zA-Z0-9\.\-\_]+(@[a-zA-Z0-9\@\.\-\_]+)?\z`. This regex matches `hola` and `hola@mullermp` but does not match `@hola`, `hola@`, and `hola@hola@ruby`.
+
 # Related Links
 
 Namespace gem discussion - https://github.com/rubygems/rfcs/issues/31
--- a/text/0008-scoped-gems.md
+++ b/text/0008-scoped-gems.md
@@ -89,7 +89,7 @@ The user can *reasonably* assume that the `hola@mullermp` gem has a namespace of
 
 An organization is a group of users that can own 0 or many gems and 0 or many gem scopes. A user can visit the `mullermp` organization by visiting the following link: `https://rubygems.org/organizations/mullermp` where `mullermp` can be any organization name. On this page, a user would see a "profile" that lists the organization's members (a user named `mullermp`), the gems owned by the organization (one gem, `hola@mullermp`), and any reserved gem scopes (`mullermp` scope). If the user is part of the organization, they can register additional gem scopes, with limitations.
 
-The organization's members are effectively the the gem owners and maintainers. On a gem's page, such as `https://rubygems.org/gems/hola@mullermp`, a new section labeled as "Organization" is displayed, with a link to the `mullermp` organization page. The organization's members, are "splat" onto the gem's page like normal gem owners.
+The organization's members are effectively the gem owners and maintainers. On a gem's page, such as `https://rubygems.org/gems/hola@mullermp`, a new section labeled as "Organization" is displayed, with a link to the `mullermp` organization page. The organization's members, are "splat" onto the gem's page like normal gem owners.
 
 # Reference-level explanation
 
--- a/text/0008-scoped-gems.md
+++ b/text/0008-scoped-gems.md
@@ -187,7 +187,7 @@ The main benefits of this approach are:
 1. It is a frugal change to the RubyGems gem specification that is mostly a naming standard. It removes validation around gem naming on `gem build` and `gem install`, and adds validation for first-time gem publishing via `gem push`.
 2. Avoids changes to the Ruby library (and its versions) and how gems are required.
 3. Similar to reserved prefixes for verified publishers (like [NuGet](https://github.com/NuGet/Home/wiki/NuGet-Package-Identity-Verification)). This approach can be considered as a *reserved suffix*.
-4. Eventual removal of "blacklisted" gem list as important standard Ruby gem names are migrated to scopes (i.e. `socket@ruby`, no need to reserve `socket`).
+4. Eventual removal of the [reserved gem list](https://github.com/rubygems/rubygems.org/blob/fc7cc3a9879d69adb48e2d37282aea34d7f311b0/lib/patterns.rb) as important standard Ruby gem names are migrated to scopes (i.e. `socket@ruby`, no need to reserve `socket`).
 
 ## Alternative 1 - Reserved Prefixes
 


indirect:
I feel like namespaces with the syntax `@scope/gemname` would be less surprising to users given that is the way NPM implements them. What are your feelings for/against that syntax?

(I also added a link to the rendered doc to the top of the PR description for easy access.)

mullermp:
The `/` portion gave issues, especially around `gem` commands, though I suppose that can be parsed correctly. Aside from that, I actually think we should NOT use the same format as it could be especially confusing when working with both languages (i.e. Rails website).

ioquatix:
If we can't get proper organisations, this might be a good first step.

I'd be a tiny bit concerned this works us into a position where supporting proper organisations is harder.

It's definitely concerning when you create a gem like `async` and then someone else clobbers your namespace by releasing a gem `async-thing-you-care-about`. Not sure if this proposal solves that problem?

mullermp:
That is valid feedback. I would think that a user's scope can be easily transferred to an organization concept in the future assuming the scoped username shares the future organization name (i.e. `s3@aws-sdk` is scoped to the `aws-sdk` user today, and then scoped to the `aws-sdk` organization tomorrow - it doesn't change how that gem is installed/required). I have an open question related to migration, that makes use of an "alias" feature. If we go that route, then to support organizations, I suppose you could even alias a gem like `s3@aws-sdk` to `s3@new-organization-for-aws` and not require the changing of code for developers either.

Though what do you mean by "proper" organizations? I assume you mean some Rails modeling that groups a bunch of owners together under some entity. Since gems can have multiple owners (already an "organization" if you will), I thought this solution fit in nicely. This proposal can certainly have a concept of such an organization entity and use the same gem naming pattern.

ioquatix:
Part of my concern is asserting that "async-foo" is official and "async-foob" is not.

mullermp:
Ah. I want to be clear that `async@foob` wouldn't be more or less "official" than `async@foo` in that example. The scope signals to developers a trusted source/organization.

mensfeld:
At first glimpse, I like this idea. It does not break too many things and seems relatively simple in implementation. My only worry is, that it won't be compatible with PURL format: https://github.com/package-url/purl-spec convention where the namespace is defined before the name:

```
scheme:type/namespace/name@version?qualifiers#subpath
```

though it could be fixed by the routing itself in the rubygems.

mullermp:
Summon @simi @hsbt, I'd love to hear your feedback on this!

simi:
I was thinking about this for long time. My biggest concern is transition time. I was thinking about some kind of backward compatible naming scheme (at least for fallback). For example if we decide to do namespacing using "rails/activerecord" scheme, we can fallback for older RubyGems/Bundler by using "rails--activerecord" (or similar, double dash is used only in few gems we can yank currently according to DB dump).

Fryguy:
Similar to that concern, I have a concern that people might squat transitional names if that's what we decide on...for example, it would be bad if I could create a rails--activerecord gem now anticipating the real gem moving to the new format

mullermp:
I think "anticipated" squatting can be mitigated (simi mentioned yanking).

mullermp:
@simi Are you suggesting scoped gems ALSO reserve Ruby namespaces? I.e. `s3@aws-sdk` MUST use an `AWS::SDK` Ruby namespace?

simi:
> Similar to that concern, I have a concern that people might squat transitional names if that's what we decide on...for example, it would be bad if I could create a rails--activerecord gem now anticipating the real gem moving to the new format

We can reject gems with `--` at RubyGems.org and add it to RubyGems specification policy as a warning.

> @simi Are you suggesting scoped gems ALSO reserve Ruby namespaces? I.e. `s3@aws-sdk` MUST use an `AWS::SDK` Ruby namespace?

No, we don't check anything in the code and I think it would be super complex to start doing that.

deivid-rodriguez:
Hi!

I like this feature proposal, and I think there's one extra benefit that hasn't been mentioned: it makes "soft-forking" easier. Say Rails wants to temporarily fork the "mail" gem, to provide a better experience until mail gem owners can get to addressing some important issues. Right now, the only way is to come up with a new name, and it's not clear how to properly communicate that the fork is only something temporary, and not meant to completely sunset the forked gem. Releasing `mail@rails` and depending on it temporarily makes this intention more clear I believe.

My main concern is namespace squatting too. I'm not sure I understand the current proposal and how do we prevent it. Can rubygems.org users create "custom scopes" (different from their usernames)? If that's the case, what prevents any random user to create the `@rails` namespace? If not, then it seems `aws-sdk` would be [already squatted](https://rubygems.org/profiles/aws-sdk)? Maybe there should be a transitional period where new scopes need to be explicitly approved to avoid this?

Regarding old clients, is the `scope--name` notation meant so that the feature works as is in old clients? I'm not sure we should choose a weird naming scheme just to support old clients. I think Bundler with a `Gemfile.lock` file would handle this pretty well since it's able to trampoline to the version that created the lockfile.

Personally, my preferred naming scheme is the one suggested by @indirect, although I understand it would require more work due to the ambiguous "/".

I'm not too sure about how to migrate to the new scheme, it seems quite complicated. I guess duplicate pushing would be best, maybe enhancing the clients to ease it, for example, something like `gem build s3@aws-sdk --alias aws-sdk-s3` that builds a "duplicated gem" with the proper legacy naming.

mullermp:
>My main concern is namespace squatting too. I'm not sure I understand the current proposal and how do we prevent it. Can rubygems.org users create "custom scopes" (different from their usernames)? If that's the case, what prevents any random user to create the @rails namespace? If not, then it seems aws-sdk would be [already squatted](https://rubygems.org/profiles/aws-sdk)? Maybe there should be a transitional period where new scopes need to be explicitly approved to avoid this?

@deivid-rodriguez FYI - I reserved `aws-sdk` user immediately prior to posting this RFC :D. I think `rails` user is also owned by the Rails team. I made the assumption that your username is your scope. I think we can certainly use an "organization" here although it requires more thought/design. An "organization" can simply be a group of users on Rubygems, who have access to 1 or more "scopes". Alternatively, the organization name can be the scope itself. I think ultimately there may need to be some initial enforcement.. some users will go and squat some names but we can root them out - I think we can reliably assume users aren't building new software on most of those published gems.

>Personally, my preferred naming scheme is the one suggested by @indirect, although I understand it would require more work due to the ambiguous "/".

Yeah, I can see reasons to want it. I started with this approach first but it introduced some complications. Specifically, we'd have to handle these cases: `No such file or directory @ rb_sysopen - @mullermp/hola-0.0.0.gem` where the gemspec's name assumes a path. Even when fixed, when it's installed, it may also create another nested directory in your gem install location, and that may or may not play nicely with existing tooling? I went down a rabbit hole and decided it was more effort than it's worth, but I could be wrong.

> I'm not too sure about how to migrate to the new scheme, it seems quite complicated. I guess duplicate pushing would be best, maybe enhancing the clients to ease it, for example, something like gem build s3@aws-sdk --alias aws-sdk-s3 that builds a "duplicated gem" with the proper legacy naming.

I'm ok with duplication as it would certainly be the safest option. In practice, as a gem maintainer with 300+ gems, it's not as feasible and probably causes some customer confusion. I think a one-way one-level alias might make sense, but we'd have to handle gem install locations too, perhaps a symlink between `s3@aws-sdk` (real source) to `aws-sdk-s3` (symlink folder).

hsbt:
👋  I'm positive to add this feature. But I'm not sure what the best syntax about `gem_name@scope` same as @indirect 

We can choose: 

* `gem_name@scope`
* `scope/gem_name`
* `@scope/gem_name`
* etc.

Does anyone summaries scoped namespace feature of other package manager?

ioquatix:
I think one of these is the most reasonable:

```
scope/gem_name
@scope/gem_name
```

the latter seems to be the format used by npm IIRC, but I think the former can be slightly better (what's the reason for/motivation of `@` character?)

indirect:
I think the motivation for the at-sign is to clearly distinguish the scope (`@scope`) from the package name (`gem_name`). It also makes it possible to talk about the scope separately from the package with the same name. For example, npm hosts a `webpacker` package in the `@rails` scope, named `@rails/webpacker`. Without the `@`, it's impossible to tell if `rails` means the scope or the top-level package `rails`.

indirect:
A tricky question about scoped packages: in Node, it is completely fine to have both `@rails/mail` and `mail` in a single project. They do not conflict.

In RubyGems, it would be (presumably) impossible to have both `@rails/mail` and `mail` in a single project, because they would both define the `Mail` constant. That means either a lot of weird errors if someone adds both packages, or a lot of extra work in RubyGems and Bundler to prevent that kind of conflict. Can a dependency on `mail` be satisfied by `@rails/mail`? If not, it's probably impossible to have a Gemfile that resolves. If yes, that's even more work that needs to be done inside RubyGems/Bundler.

ioquatix:
@indirect if I'm understanding the original proposal:

```
# @rails/mail
module Rails::Mail
```

```
# mail
module Mail
```

Is that correct?

mensfeld:
> Interesting, I have never heard of PURL. Though looking at npm's example `pkg:npm/%40angular/animation@12.3.1` it looks like `@` is escaped in the package's scope. So I imagine `s3@aws-sdk` version 1.2.3 could become `pkg:gem/s3%40aws-sdk@1.2.3`.

yes but my points was, that the notion of "scope" is not part of purl because of namespaces. However I understand the reasoning here and backwards compatibility.

mensfeld:
> Does anyone summaries scoped namespace feature of other package manager?

there's a purl schema https://github.com/package-url/purl-spec for namespaces and what I noticed is, that other registries tend to (informally) drift towards its standards.

deivid-rodriguez:
@mullermp I guess you're right, we would need to deal with squatting as just another type of possible abuse, and define clear rules about it.

@indirect I think in the particular case of `mail`, it would mostly work because most Rails users don't use `mail` directly, and `Rails` is also in control of the dependency and how it's used, so Rails would change their dependency on `mail` to `@rails/mail` and they could also choose to namespace the `Mail` in their scoped gem to stay on the safe side and avoid any conflicts with people using the `mail` gem directly.

Anyways, this small benefit is just something that came to my mind as a potential extra benefit, but it's not even mentioned in the RFC. I think this RFC only proposes a way to allow gem name collision, but does not impose anything on what top level module a given gem should define.

simi:
Another question is how to resolve dependencies? Should we use full (including namespace) gem identifier everywhere (including Gemfile.lock and gemspec dependencies specification)? I see the problem in compatibility again in here. We need to ensure transition is as smooth as possible.

deivid-rodriguez:
Yes, I think the new name should be used everywhere, otherwise it's impossible to differentiate differently scoped gems with the same name, no? I don't think there's much that can be done about compatibility, unfortunately, except for going down the route you suggested before: choosing a name scheme that's currently valid, just mostly unused. It would definitely make things smoother, although not so nice. I'm not fully sure which option is best.

To elaborate on what I said before about Bundler being able to "trampoline", the idea is that, say, support for the new naming scheme is added on Bundler 2.4. And someone upgrades to Bundler 2.4.0 to be able to bundle `@rails/mail` in their `Gemfile`. Then a `Gemfile.lock` file will be generated including the new incompatible naming scheme, and the `BUNDLED WITH 2.4.0` marker. If someone bundles this `Gemfile{.lock}` file using Bundler 2.3.0, Bundler 2.3.0 will automatically detect that the application needs Bundler 2.4 and automatically upgrade itself, so things should just work. Unfortunately this feature is only very recent and versions older than 2.3.0 would still fail hard.

simi:
@deivid-rodriguez my idea was to support both old/new name scheme (at least for some time). At RubyGems.org it could just check the client version (or we can make some additional header to ask for new scheme) and respond with given scheme.

```
gem 'rails--activerecord'
gem 'rails/activerecord' # PURL compatible if I understand it well (or any new scheme incompatible with old gem naming could be here)
```

Would work the same for some time. The latter one just will not be compatible with older RubyGems and Bundler.

_I can ask about PURL and plans of other packing repo maintainers at next OSSF meeting._

deivid-rodriguez:
Mmmm... I think we should provide a migration mechanism so that gem authors can opt in to the new naming scheme, while still being compatible with the old naming scheme. So even if the `activerecord` gem starts using `rails/activerecord` naming, plain `activerecord` in Gemfiles should still work (and by work I mean it should also pick up newer releases using the new scheme). Once users are ready (they are using up to date clients) they can start using the new scheme. And once Rails chooses to do so, it can stop providing releases with legacy naming.

I'm not really sure how we can make the above work, but if we could do it, what would be the purpose of `rails--activerecord`?

jchestershopify:
One thing I'd like to raise is implementation order of this feature vs adding teams/groups. It will be easier to implement teams and then hand out namespaces for those teams than to hand out namespaces _per user_ later on.

Otherwise there will be a landrush for _account_ names, which are largely unsupervised. Teams will need to start accounts to reserve the namespace. Once they use those accounts we lose the ability to assign responsibility for actions to individuals.

Fryguy:
> @indirect if I'm understanding the original proposal:
> 
> ```
> # @rails/mail
> module Rails::Mail
> ```
> 
> ```
> # mail
> module Mail
> ```
> 
> Is that correct?

If that's true, then this would break the "soft forking" mentioned by @deivid-rodriguez (which happens to also be my personal primary motivation for this scoping idea).  If we are expecting scoped gems to also change the internal Ruby modules, then I can't, for example, temporarily fork a transitive dependency to fix something and put it in my personal scope.

simi:
@Fryguy it is impossible (or at least I can't image how to enforce that) for RubyGems to enforce or limit any kind of module/class structure in the gem codebase. That is not the plan.

mullermp:
Based on discussions, here is where I'm at:
* It's not feasible to enforce Ruby namespace with Gem naming convention. This isn't done today for standard gems, and we shouldn't for scoped gems either. We should only encourage a standard (i.e. the [same convention in the table here](https://guides.rubygems.org/name-your-gem/).
* We probably want an "organization" RubyGems.org model that can contain 0 or many scopes and 1 or many users. Gems can either have multiple owners or ONE organization. Gems with one organization will be "owned" by the organization users (permissions modeling).
* Both `gem_name@scope` or `@scope/gem_name` patterns are probably fine, but there's preference to the latter because of familiarity (npm). Though implementing that becomes tricky.
* Perhaps also Gem specification can optionally/probably take a `scope` parameter instead of `name` being overloaded to have the pattern.

An open question I have around `@scope/gem_name` is if we should be making `@scope` folders in your gem install folder (i.e. `~/.gem/ruby/3.0.2/gems/` or equivalent). I'm not certain what breaks (is this backwards compatible?) and whether that changes how you can load a gem.

If this sounds good, I can work on updating the RFC.

mullermp:
I've pushed an update to the RFC. The highlights are:

1. Added Organizations to the design. Organizations are an entity (with a "user" profile) that have a list of members/users, list of owned gems, and a list of reserved gem scopes. I've included a light-weight design around it.
2. Added clarifications about Ruby namespaces and scoped gems. We will not enforce Ruby namespaces of any kind.
3. Addressed the migration path problem. I've worked out a neat "shim" style solution for migrating gems in place. We can't migrate gems on RubyGems.org alone, so it's on the publisher to do. The shim requires no code changes from consumers.
4. I've settled on using `gem_name@scope`. There are too many legacy concerns and downstream changes required with using `@scope/gem_name` pattern. This approach is more sane to implement and roll out. I know some people might be upset about this.
5. Decided not to use a `scope` parameter on the gem specification, it complicated implementation and usage quite a bit and provided no immediate benefits.

As always, let me know what you all think!

Also - I think this RFC is approaching a stage where I'd like to see larger scale feedback. Is there any objection to blasting this out to Ruby Weekly email series for more 👀 ? At what point do we consider RFCs to be accepted?

text/0008-scoped-gems.md
```diff
@@ -152,16 +202,25 @@ An alternative to this approach would be to reserve prefixes, like `aws-sdk-` fo
...
+  s.name = 'hola'
+  s.scope = 'mullermp'
+  s.version = '0.0.0'
+```
+This gem would still build as `hola@mullermp-0.0.0.gem` and have the same usage experience. However, would the gemspec file be named `hola@mullermp.gemspec` or `hola.gemspec`, and what implication would that have? If using `hola.gemspec`, is there a possibility someone may work with 2 `hola` gems from different organizations (x-organizational work). If the spec's name is `hola@mullermp`, we may as well overload the spec name too.
 
-**What about old versions of the `gem` command tooling?**
-Some users will be using an older rubygems version and attempt to push or install a scoped gem. What user experience do we need or want?
+**Requirements for having an organization or for reserving gem scopes**
+To prevent a rush on reservations, organizations and/or gem scopes should be limited by some initial process. Eventually this limitation would be removed. How should this be handled?
```
ioquatix, line 221: Good idea.

text/0008-scoped-gems.md
```diff
@@ -152,16 +202,25 @@ An alternative to this approach would be to reserve prefixes, like `aws-sdk-` fo
 
 In the current specification (although not documented), dots (`.`) are allowed. There are [a few “hidden” gems](https://rubygems.org/search?query=.) that use dots. If adding the `@` character is not desired, the dot character can be re-used to handle scoping, perhaps with the pattern `scope.gem_name` such as `aws-sdk.s3`.
 
+## Alternative 3 - `@scope/gem_name` pattern.
+
+While it might be preferable to use this pattern, it has implementation and legacy complications, specifically with forward slash `/`. Gem install and build will assume `@scope` as a directory and will fail. Gems installed in a location is currently flat and would have impact on existing tooling if installed within folders. Unless the `/` is escaped, the RubyGems.org API will also have issues resolving and displaying scoped gems because of Rails resource routing. This approach can be used with a more thought out design and a willingness for making large and downstream changes.
```
ioquatix, line 207: I think this should be fleshed out and seriously considered. For me it seems like short term ease of implementation vs long term pain.

text/0008-scoped-gems.md
```diff
@@ -118,26 +136,58 @@ and `ensure_gem_scope` might look something like:
...
 # Rationale and Alternatives
 
 The main benefits of this approach are:
 
 1. It is a frugal change to the RubyGems gem specification that is mostly a naming standard. It removes validation around gem naming on `gem build` and `gem install`, and adds validation for first-time gem publishing via `gem push`.
 2. Avoids changes to the Ruby library (and its versions) and how gems are required.
-3. Similar to reserved prefixes for verified publishers (like [NuGet](https://github.com/NuGet/Home/wiki/NuGet-Package-Identity-Verification)) but it can be considered as a *reserved suffix*.
-4. Possible removal of “blacklisted” gem names as important gem names can be migrated to scopes, i.e. `socket@ruby`.
+3. Similar to reserved prefixes for verified publishers (like [NuGet](https://github.com/NuGet/Home/wiki/NuGet-Package-Identity-Verification)). This approach can be considered as a *reserved suffix*.
+4. Eventual removal of "blacklisted" gem list as important standard Ruby gem names are migrated to scopes (i.e. `socket@ruby`, no need to reserve `socket`).
```
ioquatix, line 190: It's a nice idea to publish standard library gems under a scope like `ruby`.

indirect: Freeing the top level scope is a nice idea conceptually, but realistically it’s never going to happen because of the risks. Letting a rando claim the unscoped name of a standard library isn’t viable for UX and security reasons, even if it is made viable technically.

ioquatix: Ah yep fair point. I wasn't really advocating for actually freeing up existing scope, but instead ensuring that we have an "official" scope for Ruby standard library packages.

text/0008-scoped-gems.md
```diff
@@ -81,13 +83,29 @@ Hello world from my Scoped gem!
 => nil
 ```
 
+The user can *reasonably* assume that the `hola@mullermp` gem has a namespace of `Hola`, following the [recommended naming guidelines](https://guides.rubygems.org/name-your-gem/).
+
+## Organizations
+
+An organization is a group of users that can own 0 or many gems and 0 or many gem scopes. A user can visit the `mullermp` organization by visiting the following link: `https://rubygems.org/organizations/mullermp` where `mullermp` can be any organization name. On this page, a user would see a "profile" that lists the organization's members (a user named `mullermp`), the gems owned by the organization (one gem, `hola@mullermp`), and any reserved gem scopes (`mullermp` scope). If the user is part of the organization, they can register additional gem scopes, with limitations.
+
+The organization's members are effectively the the gem owners and maintainers. On a gem's page, such as `https://rubygems.org/gems/hola@mullermp`, a new section labeled as "Organization" is displayed, with a link to the `mullermp` organization page. The organization's members, are "splat" onto the gem's page like normal gem owners.
```
deivid-rodriguez, line 92:
```suggestion
The organization's members are effectively the gem owners and maintainers. On a gem's page, such as `https://rubygems.org/gems/hola@mullermp`, a new section labeled as "Organization" is displayed, with a link to the `mullermp` organization page. The organization's members, are "splat" onto the gem's page like normal gem owners.
```

text/0008-scoped-gems.md
mullermp: Other than syntax, what benefit does it bring?

text/0008-scoped-gems.md
mullermp: I think this benefit is certainly optional. My thinking was, existing Ruby standard gems can be migrated to `name@ruby` (using shim), and assuming Ruby standard gems are always pushed to `@ruby` scope, there is no need to block un-scoped equivalent. Ruby team could also register the un-scoped gem and forward it to the scoped version, too.

text/0008-scoped-gems.md
```diff
@@ -0,0 +1,229 @@
...
+An obvious drawback is, with widened gem name availability, there may be more namespace conflicts. Consider a case where org1 creates `configuration@org1` and org2 creates `configuration@org2`, both creating a `Configuration` class. If a user installs and requires both, then there may be side effects. This behavior exists today albeit gem names may be different. Developers will need to continue using discretion.
+
+# Rationale and Alternatives
+
+The main benefits of this approach are:
+
+1. It is a frugal change to the RubyGems gem specification that is mostly a naming standard. It removes validation around gem naming on `gem build` and `gem install`, and adds validation for first-time gem publishing via `gem push`.
+2. Avoids changes to the Ruby library (and its versions) and how gems are required.
+3. Similar to reserved prefixes for verified publishers (like [NuGet](https://github.com/NuGet/Home/wiki/NuGet-Package-Identity-Verification)). This approach can be considered as a *reserved suffix*.
+4. Eventual removal of "blacklisted" gem list as important standard Ruby gem names are migrated to scopes (i.e. `socket@ruby`, no need to reserve `socket`).
```
jchestershopify, line 190: I'd suggest using "denylisted" instead of "blacklisted" here.

simi: We can call it in more descriptive way. What about `blocked` or `reserved` gem list?

jchestershopify: Good call! I like "reserved" best, as it conveys intent more specifically.

text/0008-scoped-gems.md
```diff
@@ -0,0 +1,229 @@
...
+2. Help reduce [typo-squatting](https://blog.reversinglabs.com/blog/mining-for-malicious-ruby-gems) with malicious code. Scoped gems are also susceptible to typos, however, gems published under a scope are verified to be owned by that organization.
+3. An organization allows for a better gem owner permissions model. Gems owned by the same company or organization currently need to manage users for each gem. Users and gem owners can now be part of a single entity, and that entity can be the owner of the gem.
+
+# Guide-level explanation
+
+## Scoped gems
+
+A scoped gem is any gem that is named following the pattern `gem@scope`, where “gem” is the name of the gem, and “scope” is an organization's reserved gem scope. Gem scopes are globally unique across all organizations.
+
+As an example, consider a scoped gem defined as `hola@mullermp.gemspec`. Note that both the gemspec’s `name` and the required file `lib/hola@mullermp.rb` follows the `gem@scope` pattern:
```
Fryguy, line 26: Is the lib filename _requirement_ necessary?  I can see it being _recommended_, but making it a requirement makes some use cases more complicated.  For example, in the case of soft forking I'm essentially "replacing" the official gem with my copy, which also implies I don't expect callers to change the thing they are going to require.

Similarly I don't think the gemspec file name _requirement_ is necessary, but could be _recommended_, as I thought the final gem file name is produced from the `spec.name`.

text/0008-scoped-gems.md
mullermp: Sure, I'll change to "reserved". I was being explicit, because the list is called GEM_NAME_BLACKLIST defined here: https://github.com/rubygems/rubygems.org/blob/master/lib/patterns.rb.

text/0008-scoped-gems.md
mullermp: I wouldn't call them "requirements" but they are both recommended and by convention, so I would argue that you'd want to do them. The file name doesn't matter even today (you're correct that it uses `spec.name`), and the lib file's name is the same as the gem name as per the "Make your own gem" guide in RubyGems:
```
The convention is to have one Ruby file with the same name as your gem, since that gets loaded when require 'hola' is run.
```

If I understand what you mean about soft forking, let's say you want to replace `hola@mullermp` with your own implementation. You could have both `lib/hola@fryguy.rb` and `lib/hola@mullermp.rb` in your gem where `lib/hola@mullermp.rb` simply does `require 'hola@fryguy'`, which then loads the custom gem code that you've provided.

text/0008-scoped-gems.md
simi: @mullermp it was just changed to be more descriptive in there as well.

mullermp:
A question for the RubyGems team is, what level of feedback (and approval) is required for this proposal? I think the proposal in its current state is complete (pending other feedback). I'm hoping to get this pushed out to the Ruby Weekly email list for more eyes.

halostatue:
@mullermp I’m seeing this because of your reach-out, but I’m wondering why `gem@scope` rather than `scope@gem`? This would allow for (relatively) immediate namespace aliases for most existing gems. That is, `mime-types` could be aliases as `halostatue@mime-types` (this is also true of `gem@scope`, but `gem@scope` reads _backwards_ to me).

If such aliases could work, then Rubygems itself could try to load _either_:

```ruby
begin
  require full_gem_name # 'halostatue@mime-types', loading lib/halostatue@mime-types.rb
rescue LoadError
  require unscoped_gem_name # 'mime-types', loading lib/mime-types.rb
end
```

text/0008-scoped-gems.md
```diff
@@ -0,0 +1,229 @@
...
+validate :ensure_gem_scope, if: :needs_name_validation?
+```
+
+and `ensure_gem_scope` might look something like:
+
+```
+# Only invoked when needs_name_validation? is true
+# Which is on new_record? or name_changed?
+def ensure_gem_scope
+  gem, scope = name.split("@")
```
halostatue, line 137: Unless the names are guaranteed to have exactly one `@`, then there will be loss of name data.

```irb
irb(main):001:0> a, b = "a@b@c".split('@')
=> ["a", "b", "c"]
irb(main):002:0> a
=> "a"
irb(main):003:0> b
=> "b"
```

Better to force this to be strict and use `name.split("@", 2)`, and perhaps reject the scope if there are additional `@` characters in it.

halostatue (commented):
From my perspective, the scope ordering is unnatural, but it could just become Normal with enough use.

There are, I think, some suggestions for the code which may cause some potential for mayhem.

text/0008-scoped-gems.md
```diff
@@ -0,0 +1,229 @@
...
+2) Model - The Organization model will need the following fields/references: a list of owners that can perform administration (user references), a list of reserved scopes that the organization owns (string list or references), and a list of gems owned by the organization (rubygem references).
+
+3) View - The Organization view will be routed under `/organizations/<org name>`. The page will have forms for adding/removing of owners and reserving gem scopes, as well as displaying any textual information such as the organization's mission statement, a profile picture, etc.
+
+## `gem` commands
+
+To support the pattern in the `gem` commands, the [`VALID_NAME_PATTERN` regex](https://github.com/rubygems/rubygems/blob/504272e35d35a5c7dcdccfece574f5d8a060b148/lib/rubygems/specification.rb#L109) should be updated to include the `@` character (in both the specification and the policy).
+
+```
+VALID_NAME_PATTERN = /\A[a-zA-Z0-9\@\.\-\_]+\z/.freeze # :nodoc:
```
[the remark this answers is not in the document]
mullermp: Thanks, and good call out. This is mentioned in the open question section already, using `\A[a-zA-Z][a-zA-Z0-9\.\-\_]+(@[a-zA-Z0-9\@\.\-\_]+)?\z`. I don't feel strongly on what the regex (or any implementations) should look like, it was purely an example for the proposed MVP.

text/0008-scoped-gems.md
mullermp: Sure, this can be done. I just provided some sample code to get the point across, but we can definitely include better defensive checks.

mullermp:
Thanks @halostatue for the feedback. Other than implementation concerns (regex + validation), what is your general sentiment with organizations and gems having reserved scope names (effectively a gem namespaces)?

> I’m wondering why gem@scope rather than scope@gem?
I originally chose this because I tried to use npm's convention of `@scope/gem` but ran into many difficulties and incompatibilities with the gem ecosystem. I settled with `gem@scope` because scope acts like a reserved suffix. Your scope will be an organization reserved suffix. So I read this literally as "gem AT organization", rather than "organization AT gem" which is backwards. That aside, I'm open to whatever the overall community prefers.

drsharp:
Is it weird that the scoped names look like emails, sorta? the `@` symbol, especially within a string (foo@bar) is so universally understood as an email designation that it makes me wonder if a different symbol (`:` maybe?) would be better?

mullermp:
I didn't think it was weird, but I'd be open to changing it if there's more community sentiment to do so. I read it as "this gem is LOCATED AT scope", kinda like how an email would be a user name AT some website/organization.

text/0008-scoped-gems.md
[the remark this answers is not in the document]
halostatue: Ha. Yeah, I forgot to add the `+` to my regex, but my big point was the repetition of `@`.

halostatue:
The choice of scope characters is always going to be hard. `:` is not an allowable character on Windows filenames (meaning that some sort of encoding would be required for the filename), and `/` is a directory separator with its _own_ complications. As far as I know, `@` is allowed on Windows filenames. Other special characters (`~`, etc.) also have potential problems with shell substitutions on *nixish systems.

With respect to order, given your explanation (which I think would be worth adding to the RFC), I can accept `gem@scope` over `scope@gem`. It feels weird and unnatural, but that’s largely because most of the other package scoping mechanisms have started from general and headed down to specific. So Node’s `@scope/name` approach really isn’t _that_ different than `com.scope.Name` in Java. (As an aside, in `node_modules`, `@scope/name` is represented as nested directories. Not that I think that `node_modules` is in any way worth emulating.)

Regardless, if `gem@scope` is the choice made, we’ll get used to it. There is, I think a good argument for it.

It’s worth noting that, at least as of right now, neither Elixir nor Rust support scoping, and I don’t think either one is going to do so. (It could be argued that Go does, but that’s because Go uses git as a packaging mechanism.)

I think that the proposal is a good one, and I would *absolutely* want to reserve at least the `mime-types` namespace (`core@mime-types`, `data@mime-types`).

text/0008-scoped-gems.md
drsharp, line 26: Per my comment about "weird"... I just thought that: `hola@mullermp.gemspec` looked a lot like an email, especially if `gemspec` were a TLD (seems that it isn't, nor is .rb).

I'm not saying the `gem@scope` pattern isn't good (and other symbols have other problems), but just something to be aware of. 🤔

cyclotron3k:
When I download a gem named `qnap-download_station`, I can usually guess that it's going to provide a module or class called `Qnap::DownloadStation`. I know it's not guaranteed, but it's a common practice, and it's definitely the [recommended naming convention](https://guides.rubygems.org/name-your-gem/).

Adherence to this naming convention, combined with the unique constraint on gem names helps prevent/reduce namespace collisions.

So if `config@foo-company`, `config@bar-widgets`, `config@acme` all provide a `Config` class, things are going to get messy very quickly. It now becomes incumbent upon the gem developer to manually check for anyone else using that namespace?

Wouldn't it make more sense to have the owner organisation listed in the Gem's metadata instead of trying to cram it into the gem name?

djberg96:
I'm late to the conversation, but couldn't we add an `organization` attribute to `Gem::Specification`? And then modify `gem install` to allow an organization attribute?

I'm really not following the purpose of the '@' for scoping, and why the org name wouldn't be enough scoping.

I'm also trying to remember what Perl did for package management. I thought they had some way to scope them already, but my memory is fuzzy.

ioquatix:
How would I structure:

```
socketry/async -> async.gemspec -> async@socketry? async@async?
socketry/async-http -> async-http.gemspec http@async? async-http@socketry?
```

halostatue:
I believe that the proposal would have you create (to pick one of them) `async@socketry.gemspec` (although I would prefer `socketry@async.gemspec` as I do prefer `scope@gem` to `gem@scope`, despite the clear explanation provided). I would likely do `core@mime-types` (`mime-types@core`) and `data@mime-types` (`mime-types@data`) for the mime-types gems…although I’m also not entirely sure that I would stop publishing just `mime-types`.

@djberg96 I think we _should_ add an organization (or, if you prefer, _scope_) field to the gemfile, but I think that there is value in making the organization/scope part of the scoping capabilities at the command-line and in Bundler without getting excessively verbose.

I would prefer to do any one of the following

```console
$ gem install @mime-types/core
$ gem install mime-types@core
$ gem install core@mime-types
```

Over:

```console
$ gem install mime-types --organization mime-types
$ gem install core --organization mime-types
```

Similarly, I would prefer for Bundler `gem 'core@mime-types'` over `gem 'core', organization: 'mime-types'`.

In general, I think that this is a _cogent_ RFC, but it feels incomplete to me, as I think that it needs to have a clearer indication that scoped packages (either personal `@halostatue/diff-lcs` or organizational `@mime-types/core`, to use NPM-style notation) are going to end up making the transition into the gemspec and rubygems.org (**and other implementations like `geminabox`**) fairly seamlessly.

I’m not sure we should wait for _perfect_ before we act on this, *if* we act on this. I think that there are several issues here of which the scoping of gem names is only one. This gets into:

- name scoping (obviously)
- fighting typo squatting (partially)
- (partially) enabling "verified" packages (by the use of organizations)

In some ways, a lot of this (except name scoping) could be fixed if we had a *better* way of implementing signed gems (and *verifying* those signatures), rather than using namespaces and Rubygems.org as pure sources of truth. But, as someone who *did* sign gems early on and found the process excruciating (and no one verified them anyway), I don’t think that’s going to be a solution to any of these problems.

Not quite sure how to move forward on this. It’s a good RFC. I’m not sure it is *good enough*, but I also don’t know that waiting for a *better* RFC is going to give us anything this decade.

cyclotron3k:
> Similarly, I would prefer for Bundler gem 'core@mime-types' over gem 'core', organization: 'mime-types'.

But if organisation was part of the gem metadata instead of the gem name, then there would be no ambiguity in the gem name (due to the uniqueness constraint), and therefore specifying the organisation would be unnecessary, no?

zarqman:
I really like the idea of adding scopes.

Like many others, I just find that the `gem_name@scope` naming feels backwards. I think it's because we're very used to the order of `less-specific/more-specific`. Some examples:

Ruby modules: `Module::Class`
File system directories: `parent/child`
GitHub organizations/scopes: `github.com/org/project`
npm: `@scope/package`
PHP: `scope/package`
Java: `com.scope.name`

Go back in Rubygems' own history to the original gemcutter and we had `username-gemname` too.

Consider also that Ruby gems themselves often live in GitHub, where they might be named `github.com/scope/gem_name`.

Or look at the already discussed example of `aws-sdk-s3`, which exposes the `Aws::S3` namespace. While `Aws::S3` doesn't exactly match `aws-sdk-s3`, it'd be even more unnatural if the naming is reversed to `s3@aws` or `s3@aws-sdk`. Likewise for @ioquatix's example that if `async-http` becomes `http@async`, that's reversed from the included module `Async::HTTP`.

For those who care about code aesthetics and are prone to aligning related lines (including myself):
```ruby
gem 'async/async'  # feels normal
gem 'async/http'
gem 'async/websocket'
# vs
gem     'async@async'  # awkward!
gem      'http@async'
gem 'websocket@async'
```

I'd strongly prefer to see the Ruby ecosystem choose a path that's consistent with everyone else and go with `scope/name`. Or `scope.name`. Or even `scope@name`or `scope--name`. Or anything that puts the scope first.

I'll also suggest that consistency with other languages and ecosystems keeps things more accessible to new Ruby developers. If we invert the common ordering, then it's one more thing that blog posts and books will have to address. It adds mental overhead.

djberg96:
I agree that this won't prevent typo squatting since, as @andy-tycho says, you could just as well typo-squat an organization, unless we add some sort of auth hook, which is outside the scope of the project.

Where an org/scope attribute would come in handy would be for plugins that could hook into it as they see fit. Any sort of auth could be handled by the plugin author instead of us, and then users could choose to use that plugin or not.

It's not a huge deal to me either way, and it can potentially be handled by metadata of course, but it's more likely to be used IMO if it's a Specification attribute.

Fryguy:
With respect to `gem@scope` vs `scope@gem`, they are both awkward in different ways and I think we are arguing the order without also pointing out that the `@` symbol itself is part of the awkwardness.  For me, when I "read" the `@` symbol, I say "at", which implies the thing following it is a larger container (e.g. location, domain, scope, etc).  So `gem@scope` makes much more sense than `scope@gem` when only considering the `@`.  That being said the argument made by @zarqman of bigger->smaller being more consistent in Ruby is a really good one...if we go with bigger->smaller, then I think `@` is the wrong symbol choice.  Unfortunately, I'm not sure what a good symbol choice would be considering that the other expected choices of `/`, `:` have downsides as @halostatue [mentioned](https://github.com/rubygems/rfcs/pull/40#issuecomment-1118983604).

halostatue:
> > Similarly, I would prefer for Bundler gem 'core@mime-types' over gem 'core', organization: 'mime-types'.
>
> But if organisation was part of the gem metadata instead of the gem name, then there would be no ambiguity in the gem name (due to the uniqueness constraint), and therefore specifying the organisation would be unnecessary, no?

I’m not sure what you’re saying here, @cyclotron3k. I think that there is potential value in having scoping be part of a name to expand the "universe" of available package names. The points raised about how namespacing works in Ruby (e.g., when you use `mime-types`, you can generally expect that the top-level namespace is `MIME`, `Mime`, or `MimeTypes` (it’s `MIME::Types`) and that `@mime-types/core` might offer either `MIME::Types::Core` or just `Core` — but I think that this is a *social* problem to be handled).

While I don’t think that Rubygems should be enforcing any names _inside_ the gems (after all, there can be *value* in publishing a gem that monkey patches another gem to fix a bug in the original gem that hasn’t been updated…), I think that we (the Ruby community) could consider how scoped gems should expose their namespaces.

I know that I haven’t really been part of any of the previous discussions on this topic (mostly because I haven’t been aware of them), but I think that they’re fascinating. @mullermp, *thank you* for writing this RFC and getting the discussion going to a wider viewpoint.

Even though this is really a proposed Rubygems RFC, I wonder if this might not be something that should be raised on ruby-core. It’s not part of MRI implementation *as such*, but Rubygems and Bundler are so deeply part of the overall Ruby development experience at this point, it probably needs additional eyes on it.

qrush:
Hi all-

I'd like to leave an idea here that perhaps should be a separate RFC: if/when rubygems.org starts to add scoped gems support, it feels like that should be on an annual subscription service that goes directly to supporting the costs of running/maintaining rubygems.org. RubyCentral and many sponsors (see the footer on rubygems.org) covers these costs now, but this feels like a great way for the community to actively buy-in and support the central piece of the ecosystem we rely on, and especially with a feature that is geared towards organizations that have the ability to pay for services they use.

I don't know what the right amount to charge is - maybe a community survey should be held to determine that $ amount. Alternatives to gem scoping would still work fine: dashes in gem names, or a company hosting their own private gem server, so I am hoping that will be seen a reasonable workaround if an organization cannot afford the charge or refuses to pay.

Thanks all ❤️

zarqman:
Under Alternative 3, the original proposal hints at "implementation and legacy complications" arising from the use of `/` as the `scope/name` separator.

While I like the aesthetic of `/`, I do think that these complications are substantial and worth outlining more extensively.

Let's use `async-http` as our example. Further, let's assume that it would become `async/http` or `@async/http`. (I will use `~` to indicate the root directory/git repos of the gem itself.)

`async-http` presently:

1. Assumes the gemspec is located at `~/async-http.gemspec`.

2. Installs:
    `GEM_HOME/cache/async-http-1.2.3.gem`
    `GEM_HOME/gems/async-http-1.2.3/`
    `GEM_HOME/specifications/async-http-1.2.3.gemspec`
    `GEM_HOME/extensions/x86_64/3.0.0/async-http-1.2.3/` (if applicable)

3. Inside Ruby:
    Adds `GEM_HOME/gems/async-http-1.2.3/lib` to `$:`.
    `require 'async-http'` looks for `GEM_HOME/gems/async-http-1.2.3/lib/async-http.rb` by convention.

4. `gem build` outputs `~/pkg/async-http-1.2.3.gem`

Now, comparing to `async/http` and creating directories for each scope (comparable to `node_modules/`):
1. Should the `.gemspec` now be stored in `~/http.gemspec` or `~/async/http.gemspec`?
    If the latter, `gem build` would no longer be able to just search for `*.gemspec`, but would also have to include `*/*.gemspec`.

    How do either of these choices work with Gemfile commands like:
        `gem 'async/http', github: 'async/http'`
        `gem 'async/http', path: '/some/arbitrary/path'`

    It would seem that `:git` or `:github` would need to parse the scope from the `.gemspec` during install and then deliberately install into `gems/scope/gemname-x.y.z`. This might create a chicken-and-egg problem of needing to know gem.name before checking out the tree, but needing to check out the tree before reading the `.gemspec` (I'm unsure how Rubygems handles this now).

    However, for `:path`, it seems likely that the scope has to be somewhat internally discarded as the express path has already been provided. See (3) below for implications of this.

2. Now installs:
    `GEM_HOME/cache/async/http-1.2.3.gemspec`
    `GEM_HOME/gems/async/http-1.2.3/`
    `GEM_HOME/specifications/async/http-1.2.3.gemspec`
    `GEM_HOME/extensions/x86_64/3.0.0/async/http-1.2.3/` (if applicable)

3. `$:` now contains `GEM_HOME/gems/async/http-1.2.3/lib`.
    This makes `require 'async/http'` ambiguous.
    Should it be looking for `async/http.rb` inside the `async/http` gem? Or should it be looking for `async/http.rb` (yes, identical) inside the `async` gem? What happens when both exist?

    Perhaps `require` is modified to recognize scopes by prefixing the scope with `@` (same reason npm does it, I believe):
      `require 'async/http'` doesn't have `@` and so filters `$:` to only look at unscoped paths for `~/lib/async/http.rb`.
      `require '@async/http'` parses out the `@async`, filters `$:` for only `GEM_HOME/gems/async/*`, then treats it like `require 'http'` and looks for `~/lib/http.rb`.

      But, as noted in (1), use of `gem '@async/http', path: '...'` won't reliably have `scope/` as part of the pathname in `$:`, so filtering on `$:` won't work. This suggests that `$:` itself would need to be reworked, possibly as a hash: `{'@async'=>[..], nil=>[..unscoped paths..]}`. This creates backward compatibility concerns.

4. Do `gem build` and related commands output `pkg/async/http-1.2.3.gem` now? This seemingly changes how rubygems.org handles uploads, file storage, routes, etc. (as already noted in Alternative 3.)

Alternative 3 also mentions escaping `/`. I suggest that while this preserves a flat namespace, it makes a mess trying to ensure every tool gets the escaping/unescaping correct.

Further, it still doesn't resolve ambiguity with `require 'async/http'`. Should the `/` be escaped or not? If escaped, Ruby looks in `$:` for `http.rb` and hopefully finds it at `GEM_HOME/gems/async%2Fhttp-1.2.3/lib`. If not escaped, Ruby looks for `async/http.rb` and perhaps finds it in `GEM_HOME/gems/async-1.2.3/lib`.

That's a lot of change and seems to me like it would require bumping Rubygems to 4.x as it's pointing towards breaking backward compatibility. If it requires changing the behavior of `$:`, then it's also a major change within Ruby itself.

zarqman:
I agree with @halostatue that scopes should be part of the `name`, not a separate `scope` field on the spec. If there are two gems with the same name, but different scopes, a separate field creates some of the same ambiguities as organizing scoped gems into subdirectories. The only way to avoid this is to join scope+name together in nearly all usage, and if that's the case, it seems better to treat them as one from the start.

Potential places for naming collisions (with `.` as the example separator):
  `Gem.loaded_specs['async.http'] = ...`
  `~/async.http.gemspec`
  `GEM_HOME/cache/async.http-1.2.3.gem`
  `GEM_HOME/gems/async.http-1.2.3/`
  `GEM_HOME/specifications/async.http-1.2.3.gemspec`
  `GEM_HOME/extensions/x86_64/3.0.0/async.http-1.2.3/` (if applicable)
  Gem entrypoint: `lib/async.http.rb`

In all cases, if `async.` is missing, `async.http` cannot be differentiated from `faraday.http`, etc. since they'd all become simply `http`.

andrewhavens:
I think scoped gems is an interesting idea, but I'm not sure if there is a way to support it without changing the Ruby language itself. In other languages, you can import different packages with the same name and scope them within the file you are working in. In Ruby, gems are essentially globally namespaced.

The problem that was raised in this proposal about wanting to use a forked version is already achieved through the use of bundler:
```ruby
# Gemfile
gem 'mail', github: 'rails/mail'
```
This makes it clear that we are using a forked version of a gem. Thus all gems that have a dependency on `mail` will be forced to use this specific fork.

If this definition were pushed down to the gemspec level, this would make things very complicated, and even dangerous. Let's say Rails wants to depend on a forked version of a gem using something like `gem.dependency '@rails/mail'`, but what happens when another gem also has a dependency on the same gem? Does Rails get to decide that it has priority simply because it specified a specific username/org? This would open up the possibility of a gem specifying a malicious version as a dependency that takes precedence over the normal version.

So, I think this is already achieved in a reasonable way using Bundler. Might be nice to have an easy way to be able to download a gem from GitHub without having to clone it. Like `gem install @rails/mail` but that seems like a separate issue. I agree though that @username/gemname should be the format since it is the most intuitive.

ioquatix:
I think the value as a gem maintainer I see in scopes is two things:

- I want to prevent people from releasing gems into the namespace I document to my users as being official, e.g. I want to have a namespace like `socketry` and prevent other users from releasing `socketry/async-hax`
- I want to retain control over gem names within my own namespace, to avoid running into issues where I create `async-x` `async-y` and then finally find someone 10 years ago released `async-z` and won't give it up and now I have to release `async-z2` or `async-less-good-than-z`.

I've run into both of the above problems.

The biggest problem I see is:

- Top level gems are more valuable but it feels to me to be a bit of a graveyard at times with lots of good gem names essentially squatted (for some definition of: I want to use this gem X but someone last released a gem 5-10 years ago and won't give up the name.
- Nested gems like `socketry/async` don't have a clear relationship to top level names like `async`, i.e. what's the migration process? Do I use both? Release both? How does `gem install async` work when I have `socketry/async`, `async` and someone else might have created `socketey/async` (hax).

My feeling is, organisations or scopes should not change the name resolution process we already have, but instead provide a better more secure way for users to procure gems.

If we are thinking big picture, I'd suggest:

- Every user has a username specific scope, e.g. `ioquatix` -> `~ioquatix/my-gem-name` -> `my/gem/name.rb`. Everyone gets this.
- Non-username scopes, e.g. organisations / scopes, e.g. `@socketry/async` -> `async.rb`. Should be paid or gifted to major open source groups e.g. rails, rack, puma, socketry, etc.
- People can specify a resolution order in gem files, e.g.

```
source "https://rubygems.org/@rack` do
  gem "rack"
end

source "https://rubygems.org/@rails` do
  gem "rails" # -> depends on "rack" which is satisfied only by the current listed sources, e.g. [@rack, @rails]
end

source "https://rubygems.org" # general global index

gem "rando-whatever" # can pull in from [@rack, @rails, global index] in that order.
```

The good thing about this model is it allows you to fork a gem (as rails did with mail) and plug it in as a named dependency without breaking dependency resolution (because you'd need a different name to push it to rubygems.org).

indirect:
@andy-tycho that kind of "feedback" isn't okay here. consider yourself warned—next comment like that gets you blocked from the RubyGems org for at least a week.

indirect:
I don't think we should provide a scope mechanism that allows differently-named gems to provide the same global Ruby constants. I survived GitHub's original gem server, and I still have scars from trying to use an app whose gems depended on both `tenderlove-nokogiri` and `nokogiri`, which both claimed the constant Nokogiri. Bundler also can't help in that situation, because the gems have different names, different versions, and different dependency trees. In my opinion, this RFC needs a clear solution to that problem to move forward.

To me, @ioquatix's proposal to treat orgs as additional gem sources sounds like the most likely to work under the constraints we have today. For example, gems that depend on `mail` will continue to work whether `mail` comes from the global source or the `@rails` source, and there will only be one gem named `mail` claiming the `Mail` constant.

bkuhlmann:
Hey everyone. :wave:

After reading through the [RFC](https://github.com/mullermp/rfcs/blob/master/text/0008-scoped-gems.md) and this discussion, I want to add some thoughts/observations in hopes that this enriches the discussion (although I might be somewhat counter to André's concerns above -- maybe because I'm missing context to earlier days with GiHub's original gem server):

<details>
  <summary><strong>Gem Specification Scopes</strong></summary>
<br/>

In terms of gem specifications, I want to focus specifically on the use of `scope` -- as described in the [RFC](https://github.com/mullermp/rfcs/blob/master/text/0008-scoped-gems.md). I'd like to emphasize the importance of this within the gem specification as a new field:

``` ruby
Gem::Specification.new do |spec|
  spec.scope = "dry"  # This is important for many reasons which I'll highlight shortly.
  spec.name = "monads"
  # Truncated for brevity.
end
```

As the author and maintainer of [Gemsmith](https://www.alchemists.io/projects/gemsmith) -- a gem for building gems -- this would allow organizations and individual contributors to configure this information once via Gemsmith's [XDG](https://www.alchemists.io/projects/xdg) configuration. This equates to being able to build a gem as follows:

``` bash
# Uses global scope as exists today or pulls scope from XDG configuration (if configured).
# This is a nice productivity boost when building multiple gems within the same scope.
gemsmith --build demo

# Uses custom scope which overrides any XDG configured local or global scope.
# Definitely tedious when creating multiple gems within the same scope -- if not using an XDG configuration -- but handy for one-time overrides.
gemsmith --build monads --scope dry
```

This also means that Gemsmith -- and Bundler -- wouldn't have to add special logic for parsing a gem name -- at creation -- by splitting `dry@monads` into `dry (gem scope)` and `monads (gem name)`. Even better, we improve the developer experience for creating new gems by not forcing someone to have to type this:

``` bash
gemsmith --build dry@monads
```
</details>

<details>
  <summary><strong>At (@) Symbol Avoidance</strong></summary>
<br/>

Building upon what I've demonstrated above, I'd like to push for avoiding the use of the at symbol (@) within the gem name, package, and URL altogether for the following reasons:

- Use of `<gem>@<scope>` is backwards, awkward, and not intuitive which many have pointed out already.
- Use of `<scope>@<gem>` is better but -- as **Dan** pointed out earlier -- feels more like an email address which confuses me as well.
- As **Maciej** mentioned earlier, use of `@` is not entirely compatible with the [Package URL Specification](https://github.com/package-url/purl-spec) and would best be reserved for version information.
- Use of `@` in the pathname also feels awkward and non-intuitive to me. Example: `$HOME/<truncated>/3.1.2/lib/ruby/gems/3.1.0/gems/dry@monads-1.4.0`.
- Use of `@` in the URL doesn't make sense either. Example: `https://rubygems.org/gems/dry@monads`

I'd like to suggest as an alternative which is to keep scope information in the gem specification. Then both Bundler and RubyGems would be able to do the following:

**Paths (gem installation and management)**

``` bash
# Global scope as exists today.
$HOME/<truncated>/3.1.2/lib/ruby/gems/3.1.0/gems/dry-monads-1.4.0

# Scoped as being proposed.
# NOTE: `@` is removed in favor of using `dry` as a scoped directory structure.
$HOME/<truncated>/3.1.2/lib/ruby/gems/3.1.0/gems/dry/monads-1.4.0
```

:warning: There are definitely complications with this approach that I'm glossing over as **Thomas** has detailed [here](https://github.com/rubygems/rfcs/pull/40#issuecomment-1119921834) but I think they are surmountable.

**URLs (gem lookup)**

``` 
# Global scope as exists today.
https://rubygems.org/gems/dry-monads

# Scoped as being proposed.
# NOTE: `@` is removed in favor of using `dry` as a scoped directory structure.
https://rubygems.org/gems/dry/monads
```

ℹ️ In all of the above use cases -- and as emphasized in the RFC -- the gem namespace would remain the same regardless of using global or specialized scope. Example:

``` ruby
module Dry
  module Monads
  end
end
```

The only difference is how Bundler finds and resolves the gem locally (i.e. either using the scope if defined or falling back to global if not) and how RubyGems lists the gem in the URL (which also depends upon the gem specification).

</details>

<details>
  <summary><strong>Graceful Degradation, Soft Forking, and Migration</strong></summary>
<br/>

So far everything I've been proposing allows for graceful degradation, soft forking, and gem transition support. By this, I mean gems can exist as they are today with support for scoped coexistence while falling back to the existing and established format. To summarize:

**Gem Specification**

``` ruby
# Valid
Gem::Specification.new do |spec|
  spec.name = "dry-monads"
  # Truncated for brevity.
end

# Valid
Gem::Specification.new do |spec|
  spec.scope = "dry"
  spec.name = "monads"
  # Truncated for brevity.
end
```

**Paths**

``` bash
# Valid (global)
$HOME/<truncated>/3.1.2/lib/ruby/gems/3.1.0/gems/dry-monads-1.4.0

# Valid (scoped)
$HOME/<truncated>/3.1.2/lib/ruby/gems/3.1.0/gems/dry/monads-1.4.0
```

:warning: Keep in mind two formats of the same gem version *would not be allowed*. I'm only showing the same version for path comparison purposes.

**URLs**

```
# Valid (global)
https://rubygems.org/gems/dry-monads

# Valid (scoped)
https://rubygems.org/gems/dry/monads
```

All of this means that you can do the following:

- Gracefully degrade to global scope if a custom scope isn't provided.
- Allow gems to be soft forked by using `my_scope/monads` as a temporary quick fix while the main gem catches up.
- Allow existing gems to migrate to the new scoped format by releasing a new version which adds the `scope` to their gemspec.
</details>

None of what I've written above addresses the name squatting problem, though. That is still a complication which has been mentioned in this discussion but probably warrants a different proposal.

indirect:
@bkuhlmann I think your suggestion is aligned on the end goals: scopes need a way to avoid global namespace conflicts. 👍🏻 I might have missed it, but l didn’t see anything in your post to address “scoped forks”, like the Rails org creating their own Mail gem that is an alternative/replacement for the global Mail gem. How would you handle that?

indirect:
@andy-tycho Baseless rhetorical claims that the problem is rare (which are empirically false, by the way) is not the kind of feedback we are looking for. An entire GitHub account with a fake profile picture just so you can post contentless angry negative comments about this RFC is not acceptable behavior. You were warned, and now you're out.

schmijos:
> I'm late to the conversation, but couldn't we add an `organization` attribute to `Gem::Specification`? And then modify `gem install` to allow an organization attribute?
> 
> I'm really not following the purpose of the '@' for scoping, and why the org name wouldn't be enough scoping.

@halostatue I agree with @djberg96 mainly for the reason that we already have got possibilities to "scope" gems. We can already do `git`, `github`, `ref`, `branch` and whatsoever.

mullermp:
>I must as well note that I am fairly surprised that such a discussion happens without @matz in sight. And @dhh. And @ko1. And @tenderlove, @hone, @amatsuda, @tmm1, @kobaltz, and all the other prominent locomotives and shapers of the Ruby world.

I would love for the "prominent locomotives and shapers of the Ruby world" to comment and contribute to this RFC!

kobaltz:
I think that this could be a good change for the future of Ruby as a whole. My main concern is backwards compatibility with existing applications. Though, they're probably running an older version of Ruby and rubygems anyways, so it likely wouldn't matter as long as the API was backwards compatible.

Ultimately, what is the goal of scoped gems and what problem is it solving? Based on the conversations above, it looks clear that the scoped gems is providing "confidence" that a consumed gem is from a certain organization. If this is the goal, then sure, it is moving in the right direction.

However, there was other mentions of consuming potentially malicious gems and squatting on names. Sure, this would help combat the squatting on gem names as things are now scoped. However, I'd push back on the malicious gems bit. Someone could create a fake org scope like `hotewire` instead of `hotwired` and then publish something malicious there. At a glance, it may look legit. If this is a main reason for the scoped gems, I don't think it will solve the problem that it is aiming to solve. Although, if there is a requirement for an organization to be verified in order to gain access to the scoped gems, then we may be on track to having more legitimacy to the new convention.

As far as the naming convention. I'll go with the flow. I don't have a preference on gem@scope or scope@gem or @scope/gem. gem@scope does give a more normal feel as it is similar to user@server.

mullermp:
I want to thank everyone for providing their feedback and perspectives on this. I think the next step here is to consolidate/parse the feedback and determine what changes are needed. At a glance, the idea seems to be overwhelmingly positive, but the approach (naming and usage) is mixed. I understand that we can't please everyone. Hopefully we can strike a happy medium here.

bkuhlmann:
> **André**: l didn’t see anything in your post to address “scoped forks”, like the Rails org creating their own Mail gem that is an alternative/replacement for the global Mail gem. How would you handle that?

Yeah, fair point. I don't address that very well and I'm not sure I have a good answer other than what I [commented on earlier](https://github.com/rubygems/rfcs/pull/40#issuecomment-1120448864) and what **Samuel** [mentions in his comment](https://github.com/rubygems/rfcs/pull/40#issuecomment-1120030385) (i.e. the Rails Mail gem example) without thinking through the directory and URL path design a bit more (as well as eliciting more feedback). I agree there are caveats to think through and address better. Something that would be a *huge help* is to see the RFC be brought up-to-date with the current discussion so, at a high level, everyone is back on the same page and can help progress the design even further.

**Matt**: Maybe you can update your [RFC](https://github.com/mullermp/rfcs/blob/master/text/0008-scoped-gems.md) -- if you are not already in the process of doing this -- to detail the directory path design as mentioned in these comments and discussion? If your RFC was brought up-to-date with the current discussion then it'd be easier to iterate on this a bit more?

ioquatix:
Would it be helpful for me to write a RFC for the proposal I outlined too? Even if it's just as a counter point?

midnight-wonderer:
@indirect Regarding scoped forks, the issue is very similar to Linux packaging.  

The solution is:  
Each Linux package can declare `provides`, `requires`, `conflicts`, `obsoletes`, `recommends`, and `supplements`.

For example, the `java-21-openjdk` package in my machine declared `provides` as follows:
~~~
Information for package java-21-openjdk:
----------------------------------------
Repository     : Main Repository (OSS)
Name           : java-21-openjdk
Version        : 21.0.6.0-1.1
Arch           : x86_64
Vendor         : openSUSE
Installed Size : 523.6 KiB
Installed      : Yes (automatically)
Status         : up-to-date
Source package : java-21-openjdk-21.0.6.0-1.1.src
Upstream URL   : https://openjdk.java.net/
Summary        : OpenJDK 21 Runtime Environment
Description    : 
    The OpenJDK 21 runtime environment.
Provides       : [28]
    libjawt.so()(64bit)
    libawt_xawt.so()(64bit)
    jre1.3.x
    jre1.4.x
    jre1.5.x
    jre1.6.x
    jre1.7.x
    jre1.8.x
    libsplashscreen.so()(64bit)
    jre1.10.x
    jre1.9.x
    java-21-64
    jre-21-64
    java-21-openjdk = 21.0.6.0-1.1
    java = 21
    java-21 = 21.0.6.0-1.1
    java-64 = 21
    java-fonts = 21.0.6.0
    java-openjdk = 21.0.6.0-1.1
    java-openjdk-64 = 21.0.6.0-1.1
    jre = 21
    jre-21 = 21.0.6.0-1.1
    jre-21-openjdk = 21.0.6.0-1.1
    jre-21-openjdk-64 = 21.0.6.0-1.1
    jre-64 = 21
    jre-openjdk = 21.0.6.0-1.1
    jre-openjdk-64 = 21.0.6.0-1.1
    java-21-openjdk(x86-64) = 21.0.6.0-1.1
~~~

When more than one packages provide the same functionality, the end user has to select which one to use.

I'm not sure if we can adopt something similar. If we can, in the case of forking `nokogiri`, it would involve creating a pull request to the original repository to update the `gemspec` with the necessary information.
(And also significant tooling upgrades on the bundler side.)

If this is agreeable, the `conflicts` & `provides` functionality can be resolved as a separate matter from the scoped name being discussed here. Potentially this should be discussed together with https://github.com/rubygems/rubygems/issues/2731#issuecomment-1328144005.

simi:
@midnight-wonderer linux package managers can usually have only 1 package with 1 version installed at the time. But RubyGems can host multiple versions together. Even it seems similar, IMHO it is really not similar environment.

midnight-wonderer:
Thank you for the feedback; I just realized I wasn't being clear.

What I meant to say is that if it’s agreed the module conflict issue is separate from the scoped gem, we should discuss this separately. I did not mean to imply agreement with the Linux solution itself.

To elaborate: if there is a solution to the problem, I don't think the solution would affect this RFC nor what is proposed here can be a part of the solution.

Kerrick:
I noticed this RFC got closed today. Did the following happen?

> I want to thank everyone for providing their feedback and perspectives on this. I think the next step here is to consolidate/parse the feedback and determine what changes are needed. At a glance, the idea seems to be overwhelmingly positive, but the approach (naming and usage) is mixed. I understand that we can't please everyone. Hopefully we can strike a happy medium here.

If so, I cannot find the new RFC.

indirect:
Hi Kerrick. I don't think Matt intended to close this RFC, just delete his fork of this repo.

That said, I do not expect this RFC to go anywhere--the RubyGems GitHub organization and gems were recently [stolen from their maintainers of the last 15 years](https://joel.drapper.me/p/rubygems-takeover/), so the team involved in creating and discussing this RFC no longer works on RubyGems.

mullermp:
Yes, I deleted my fork. I would consider reopening this despite circumstances if possible because it's still a useful feature request.

PR #44, opened 2023-01-05
Gem Contents Ingestion Proposal

martinemde:
RFC for serving gem contents.

[Rendered](https://github.com/rubygems/rfcs/blob/martinemde/gem-contents-api/text/0008-ingest-gem-content.md)


indirect:
This seems like a great starting point. If we are worried about the total S3 storage costs, we could set the CDN expiration very high, set the rate limit very low, use robots.txt to exclude scrapers, and only supply the contents on demand via a lambda or equivalent.

I believe that’s the reason that NPM supplies the files addressed by content hash—you get full deduplication for free, since any file that’s the same between versions has the same content hash and can be reused without taking up any more space. Might be worth considering for our system, or maybe we do that in the backend while Fastly caches the content under a URL that matches the filename?

text/0008-gem-contents-api.md
```diff
@@ -155,28 +155,20 @@ If there is a way to create a subset of gems that covers an acceptable percentag
 The S3 bucket should be exclusively for expanded gem contents.
 The contents of a gem will never change, so the storage should be treated as read-only except to remove yanked gems.
 
-It will be necessary to store a manifest that indexes all the files in the gem, their size, checksum, and other information. While file contents will never change, the calculated information inside the manifest may need to be updated as our information about the usage of this feature improves or other needs arise. Making this information queryable will improve the feature. Therefore, it makes sense to use the database to store the manifest information.
+It will be necessary to store a manifest of all the files in the gem. The manifest will return a list of all files with all the necessary information about each file: content type, size, path, checksum, binary, etc.
 
+Rubygems.org can show a placeholder for the content browsing page until a gem has its contents stored..
```
olleolleolle, line 156:
```suggestion
Rubygems.org can show a placeholder for the content browsing page until a gem has its contents stored.
```

simi:
Feel free to ping me once this is ready for comments. I have some initial questions and notes to share.

indirect:
This looks great to me, and I think we could probably bring it out of draft to get broader comments, if you're happy with it.

I'm not 100% sure about a separate database row for every file in every gem. Could we get away with a jsonb field and one row per gem version? Even if we do that, won't that mean many copies of information about repeated files, like the filename and the content hash over and over? Are there other options that would be less repetitive?

I don't actually have a better idea of the top of my head, but I thought the question was worth asking in case someone else did have ideas. 😄

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,337 @@
...
+### Yanking gems
+
+The hashed filenames pose a slight challenge when yanking a gem.
+The relative infrequency of yanking means that we can accept a more involved process removing files than adding them or hosting them.
+
+When a version is yanked, the file manifest for that version will be used to search for any hashed files that are unique to the yanked gem.
+If a SHA256 named file would be orphaned by yanking the gem (it has no gem versions that reference it) then the file will be deleted from S3 and purged from Fastly's cache.
+If a file has other references, then the file was already a duplicate of a file in a previous version. Files with multiple references will be preserved until the last reference is removed.
+
+When a gem is yanked the manifest table rows recording the file names with checksum, size could be preserved but no longer shown anywhere.
```
alightergreen, line 92: Should there be a "status" or "yanked T/F?" column in the table where we can deterministically record if a file was yanked? (Or does Ruby have a nice "soft delete" feature built in we can coopt for this purpose?) 

Since "a record with no source file" would under normal operations be considered a "bad" or "error" state, I think it's worth tracking when we MEANT to get something into that state so we can differentiate between actual errors and intentional actions.

alightergreen: Ah! I see you had a similar thought below. :D

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,337 @@
...
+- What is the impact of not doing this?
+
+# Unresolved questions
+
+- Feedback on naming. We can tackle column names in individual PRs, but if there is any structural names we'd like to refine, let's address them here.
+- Should we keep the file manifest of yanked gems?
+  Is a yanked gem intended to be trashed completely or just hidden?
+  One possible downside to keeping the file manifest would be if files were committed that leaked private info just by their name or content type.
+  I'm leaning towards not saving yanked file manifests, but happy to hear comments.
+- Should we try to trigger gem ingestion on-demand when a manifest is requested for a gem that doesn't have one yet?
```
alightergreen, line 318: I think we should give people a button they have to click to say "yeah I want this please go unpack it for me I understand it might take an hour". 

I think that would avoid the web crawler problem, right? And also a "I'ma just clickin around, checkin things out" person from launching 100 background jobs on stuff they never intend to come back for.

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+- Start Date: 2023-01-18
+- RFC PR: (leave this empty)
+- Bundler Issue: (leave this empty)
+
+# Summary
+
+Create a process for the purpose of processing and saving the contents of a gem.
+The process will run each time a gem is uploaded to save information about each file in the gem.
+The saved output will include file size, content type, SHA256 checksum, and relative path within the gem.
+In the interest of simplifying this RFC and the feature work that follows, this RFC focuses on only the ingestion and storage of gem content.
```
segiddins, line 11: I think it's worth also include gem metadata, specifically the gem spec file (calling it out because the canonical gem spec is stored in a separate tarball than the rest of the gem content)

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+
+# Motivation
+
+It is our hope that we can provide access to individual file data from gem files across versions in a compact and efficient way that supports future features depending on gem content.
+
+Creating a gem content ingestion and storage system will enable rubygems.org to support the security and useability gems in the Ruby Gems ecosystem.
+The gem content repository will be the authoritative source for the content of gems, accessible as individual files without without requiring downloading and unpacking a gem.
```
segiddins, line 18: nit: it isn't actually authoritative, the .gem files remain the source of truth (since that's what will actually get installed)

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+
+In the near future, the indexed gem content will allow an API for accessing the gem content.
+The API will support both rubygems.org and external projects that wish to use individual gem content.
+Rubygems.org plans to add tools for browsing gem content on rubygems.org or through the `gem` command line interface.
+
+# Guide-level explanation
+
+For the purpose of illustration, I will use the gem `rake` as an example.
+
+When a gem is uploaded, rubygems.org saves it as a tarball file named `rake-13.0.6.gem`.
```
segiddins, line 28: nit: this is a `gem version` in rubygems.org terminology

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+Binary files will be recorded in metadata, but not stored.
+Gem contents will be stored at an S3 key named for the SHA256 of the file to de-duplicate files within a gem.
+Yanking will remove file contents unique to the gem, as expected, without altering unyanked gem contents.
+
+### Background process
+
+In order to ingest the contents of a gem, a background process can be enqueued to index and store the gem contents.
+
+At gem upload time, or on demand, a background worker will download, unpack, record file metadata, index files in the database as rows, and upload each file to a content-only S3 bucket.
+The process will result in a row in the database for each file in a gem, with the relative path, checksum, and size of every file.
```
segiddins, line 42: instead of a raw checksum, we should store an SRI string here, to support seamlessly migrating checksum algorithms in the future

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+
+### Indexing files to create a manifest
+
+The background worker will create a json document manifest of all the files and store it in S3 alongside the files themselves.
+
+The manifest will hold the path, checksum, size, content type, binary determination, and plain text line count.
+
+In order to support yanking gems, which requires comparing checksums across all versions of a gem, a table will be added to the rubygems.org database.
+The table will consist of one row per gem version with a foreign key for the `version_id` and a jsonb column holding only the SHA256 checksums of all files in that version.
```
segiddins, line 51: why one row per version, and not per version per file? I think if each version/file had its own row, it would make garbage collection much easier, since for each file as we delete it, we can check to see if that was the only reference with the same checksum

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+
+The table will be indexed on `version_id` to support looking up all checksums in all versions of a gem.
+This table is for the sole purpose of yanking a gem, discussed below.
+
+### Recording, but not storing binary files
+
+Files that are identified as binary will be flagged in the index and not uploaded to the storage bucket.
```
segiddins, line 58: how will we identify binary files?

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+
+As of this writing, there is more than 5TB of gem files stored by rubygems.org.
+
+Using `rake` as an example, the total size of all `.gem` files is currently 6.8MB.
+
+The average compression across the entire history of the rake gem is about 84%.
+Expanded fully, all versions of rake add up to 42MB.
+
+Many files in rake do not change between versions, allowing us to optimize the storage.
+If each file is instead saved at its SHA256 checksum, causing and identical file to be stored only once, the size of all uncompressed files in `rake` droms from 42MB to 10MB.
```
segiddins, line 73:
```suggestion
If each file is instead saved at its SHA256 checksum, causing and identical file to be stored only once, the size of all uncompressed files in `rake` drops from 42MB to 10MB.
```

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+The resulting directory has 1226 files for all versions of `rake`.
+
+In order to access a hashed filename, access the file content table to find all files in the version of rake.
+Use the relative file path within the gem to find the SHA256 checksum.
+Access the S3 bucket at a key similar to `/rake/files/c588eda0547f6fa72fd7cd9ba2ddc2e81b96e61d690a929e8f579412da27eebd`
```
segiddins, line 78: should the key include the gem name? we might achieve significantly better de-duplication if we only key on the content and nothing else, since I'd imagine there are some files duplicated across gems

text/0008-ingest-gem-content.md
segiddins: I think we'd be able to check the yank status by looking at the referenced gem version row

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+
+When a version is yanked, the file manifest for that version will be used to search for any hashed files that are unique to the yanked gem.
+If a SHA256 named file would be orphaned by yanking the gem (it has no gem versions that reference it) then the file will be deleted from S3 and purged from Fastly's cache.
+If a file has other references, then the file was already a duplicate of a file in a previous version. Files with multiple references will be preserved until the last reference is removed.
+
+When a gem is yanked the manifest table rows recording the file names with checksum, size could be preserved but no longer shown anywhere.
+
+### Indexing older gem versions
+
+Depending on the implementation of future fetaures, gems can be indexed on demand by request or in an automated way.
```
segiddins, line 93:
```suggestion
Depending on the implementation of future features, gems can be indexed on demand by request or in an automated way.
```

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+  yanked_at (optional, see unresolved questions)
+  ```
+9. New text files are uploaded to a gem contents s3 bucket with key: `/(gemname)/files/(sha256)`
+  (The content S3 bucket is checked to see if a file already exists at the key before uploading.)
+10. An attribute is updated on the Version record to indicate the gem content has been successfully uploaded.
+  (This is necessary because a partially ingested gem would show a partial file list. Better to show all or nothing.)
+
+### Yank process
+
+When a gem is yanked, additional steps will be performed to remove files and stop serving manifests.
```
segiddins, line 134: we will likely need to also update the fastly purging in the yank codepath to account for the new files being deleted

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+# Drawbacks
+
+### Storage and Processing Costs
+
+The most obvious drawback is storage and processing costs.
+The increased storage on S3 will cost more.
```
segiddins, line 151: how much? can we estimate what this will cost, based on total anticipated data storage volumes?

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+
+# Guide-level explanation
+
+For the purpose of illustration, I will use the gem `rake` as an example.
+
+When a gem is uploaded, rubygems.org saves it as a tarball file named `rake-13.0.6.gem`.
+A gem may be installed with `gem install` or the `.gem` file can be downloaded using `gem fetch`.
+Downloading the entire gem locally is currently the only way to access the contents of gem.
+
+With the gem content ingestion system in place, gem contents will be saved to S3 and indexed in the rubygems.org database.
```
segiddins, line 32: will we use a new S3 bucket for this purpose? (that would make sense to me)

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+#### Global Checksums
+
+For a time we considered applying this checksum approach globally.
+The apparent benefit would be in deduplicating files across gems.
+Files like LICENSE and other boilerplate could be deduplicated across the entire gem repository.
+Some gems vendor other gems to reduce dependencies (e.g. bundler).
+It seems like there could be enough benefit to warrant this solution.
+
+However, we decided against it for a few reasons, the primary being yanking gems.
+If we had to consider every file as a possible orphan, we would have to search the entire file table without constraints for a matching checksum before deleting a file.
```
segiddins, line 253: i dont think this holds if we store each file in its own row

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+If we instead constrain our deduplication to only files within the same gem, we get most of the benefit without the increase in file overlaps.
+When only files within the same gem are considered for deduplication, we may miss a few small optimizations, but we reduce complexity.
+Since the files within a gem can only change with versions of that gem, there's less likelihood for race conditions.
+If one gem adds a file, while another yanks the only other reference to that file, we could end up with the incorrect state.
```
segiddins, line 257: not necessarily, if we have a database table that holds the SHAs, we could use transaction locking against that row to serialize changes & avoid this race

text/0008-ingest-gem-content.md
```diff
@@ -0,0 +1,332 @@
...
+Not storing the only binary file in this gem saves 99% of the uncompressed storage space while providing almost all the benefit of offering gem browsing.
+
+A user that wants to examine a binary file in a gem is effectively downloading more than the file size of the gem.
+They would be better served downloading the gem for that purpose.
+
+#### Detecting Binary Files
+
+There are many solutions that can be found for detecting binary files.
+We intend at this time to use `grep` to quickly print a list of files that it thinks are binary, and therefore not text searchable.
+We do this by relying on `grep` to output when a binary file matches.
```
segiddins, line 288: I think this might be significantly slower than using https://rubygems.org/gems/ruby-magic, since it's spawning a subprocess?

text/0008-ingest-gem-content.md
indirect: I’m worried that one row per version per version per file will produce a table with hundreds of millions of rows and potentially congest the primary database—though I’m open to other suggestions about how to avoid that.

text/0008-ingest-gem-content.md
martinemde: If I understand correctly, you mean pulling the metadata.gz file out of the gem and hosting that along with the gem's content? That makes sense to me.

text/0008-ingest-gem-content.md
martinemde: At the risk of nitpicking semantics, I meant "authoritative" because rubygems.org is the authority on the content of any given gem. It may not be the definitive source, since the definition of a gem content is "what is in the gem file" but we certainly can say, with authority, that the content we're representing is coming "directly from the authority on rubygems.". I'm looking for a word that we could agree on: verifiable source? authentic source? dependable source? I think authoritative captures it since we're certifying the content is correct by proof of authority over the gems.

text/0008-ingest-gem-content.md
martinemde: fixed, thanks. Knowing it was a tarball was useful for me, but doesn't need to go in the RFC.

text/0008-ingest-gem-content.md
martinemde: Yes, this was a compromise to try to reduce the number of rows. The rows will never be used individually, but only as a group for each gem version. Some rough pseudo code to yank a version would look like:
```
keep_files = load_all_checksums_excluding(yanked_version)
yanked_files = checksums_for(yanked_version)
remove_files = yanked_files - keep_files # remove any files that exist in another version.
storage.delete_files(remove_files)
delete_version_checksums(yanked_version)
```
This is the only interaction where we would use this table.

text/0008-ingest-gem-content.md
martinemde: The idea I like best is using grep, since the algorithm for determining which files are "binary" will match with most developers experience of files for which grep doesn't return results.

text/0008-ingest-gem-content.md
segiddins: hmmmm....

I was thinking we'd use the rows to enable redirects:

`/gems/rake/versions/13.0.0/content/lib/rake.rb` -> `s3 backed domain/gems/rake/$HASH` ?

text/0008-ingest-gem-content.md
martinemde: I discussed this in the rationale section. Briefly: the deduplication benefit requires all files in all gems be searched for duplicates when deleting any gem. This requires one row per file, which was my original proposal. The _only_ benefit that I can see in one-row-per-file is this minor deduplication. I expect it to be a tiny benefit since it will only deduplicate gems vendored in other gems (I expect this to be a very small portion of total gems) and files that are frequently the same, like LICENSE.

I also think it significantly increases the yank/push race condition situation since now any gem, across all gems, could trigger the race instead of just versions within the same gem (which I still don't think is insignificant, but at least it's a small enough number of files that we can verify all of a gem's content files after it's yanked).

text/0008-ingest-gem-content.md
martinemde: I'll add mention of that. Thank you!

text/0008-ingest-gem-content.md
martinemde: Agreed. I will try to make that more clear since that is what I had in mind.

text/0008-ingest-gem-content.md
martinemde: I don't have an estimate of timing. The benefit of grep is that it would return all binary files in the response of one subprocess. However, I still won't have content-type which may mean using `ruby-magic` rather than `file`. As I started prototyping this to understand it better, I ran into the problem where checksum, file type, and `binary?` all needed to load the file into memory in a different way. Maybe I could put in here all the options we can think of, and then leave it as an open question to benchmark the different approaches. `ruby-magic` is on my radar to try, in particular because `file` may vary in output between *nix systems.

text/0008-ingest-gem-content.md
martinemde: Rather than reply to another related comment below about one row per file, I will attach here my response to your comment below:

I have a part of the document I haven't committed yet that tries to deal with this particular race condition. I'll upload it soon (it's just a little messy, working on that today). It's relevant whether we group by gems or globally. I'm not sure locking individual file rows would be sufficient, though I haven't been able to completely solve the race condition yet without just "re-checking" s3 at the end of a process and re-entering the entire process if anything is incorrect. I'd be happy to hear your thoughts on how to approach this.

I'm also curious if you have a different view of the benefit of the additional deduplication. If I'm wrong about how much space it would save, it could be worth the additional database load and size. I'd hate to simply trade saving a few 100MB on S3 for adding few 100MB index to the postgres database. The only purpose of the index is for yanking gems without a 100MM row table scan, and yet the index could arguably be larger than the savings on S3. Please do correct my if you know of a more significant deduplication savings than I'm estimating that could counteract that. 

Originally I had planned to store a row per file for every file that included all of the information, like content-type and size. That does not optimize for loading the manifest which is vastly more common than yanking gems. We rewrote the proposal for storing the manifest statically on S3 and only storing the hashes necessary for yanks in the database. It's even arguable that yank is uncommon enough that we could do the entire process by loading all the static manifests from S3 for every gem (they'll all be in a single directory within the gem). Then any locking or race condition solution would be entirely managed through S3. I haven't thought it through yet, but I don't see why that wouldn't be possible.

text/0008-ingest-gem-content.md
martinemde: That redirect pattern is one of the benefits.

On the contrary, unless you're already viewing the source somewhere else, knowing that `lib/rake.rb` exists at all means you already requested the manifest and you therefore have all the data for mapping path to hash.

text/0008-ingest-gem-content.md
indirect: The current bucket is 4.8TB, growing about 100GB/month. Estimating that we would need 1.5x that much space for content-deduped but not compressed files, the bucket would need to host around 7.2TB. Using the AWS calculator, that would increase our current AWS bill by roughly 5-7%, which I think is acceptable.

text/0008-ingest-gem-content.md
martinemde: I can see why that would be nice.

Let me just play devils advocate for a moment (though I don't object to SRI in theory). Migrating would never be all at once. We'd have to re-digest all files to continue to benefit from the deduplication, and that would bust all caches for any gem. The algorithm itself isn't really for the purpose of verifying the file content (though that's an added benefit). It's real purpose is a deterministic algorithm that quickly helps us find duplicates. Maybe in 10 years I'll eat my words when there's a painful algorithm migration in rubygems.org.

Since this a naming convention rather than a critical integrity check I'm not sure we'll need to migrate. If we want to add integrity checks to the manifest later, we can use an additional attribute on each file and continue with our naming.

The only concern that would make me not use SHA256 right now is if there was a credible likelihood of multiple files having the same checksum but different fully valid content within the same gem (or multiple gems if we use global hashes).

text/0008-ingest-gem-content.md
indirect: The more I think about this the more I am leaning towards a second RDS Postgres that hold just the table of one-row-per-file-per-version? It's somewhat higher infra complexity to have a second postgres, but I think it's much lower codebase and operational complexity because the contents API loses the ability to take down any main site functionality.

That way we can index on both path and content hash, get single row results immediately, and never worry about congesting the main database. Rails 6/7 has excellent support for models that talk to a different database, and if it's in a separate postgres instance I would feel completely comfortable with a single table with hundreds of millions of rows.

text/0008-ingest-gem-content.md
martinemde: Responded [above](https://github.com/rubygems/rfcs/pull/44#discussion_r1085998314) to keep the discussion in one place.

text/0008-ingest-gem-content.md
martinemde: Responded [above](https://github.com/rubygems/rfcs/pull/44#discussion_r1085998314) to keep the discussion in one place.

text/0008-ingest-gem-content.md
martinemde: @indirect Is the purpose of this to be able to load a file by a nice path rather than a hash? We could arguably tell S3 to configure this bucket for static web hosting, and then add [static redirects](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-page-redirect.html) for all file paths to the appropriate hash.

martinemde: Between json blob and rows with only version and sha, I think the size will be about the same, but if you add the necessity of an index on the global hashing, one row per file, you will add that large index that is not needed in the per gem approach.

If you want path and want to perform redirects, then I would not use the json blob approach. This would also change this table to be significantly more read heavy than write heavy.

indirect: I think ~100,000,000 static redirects for a single S3 bucket is... probably too many? I haven't checked on the limits, though.

martinemde: Personally I don't think the pretty urls are worth the extra complexity since almost every request ever will use the eventual website user interface which _must_ load the manifest first and will automatically hide any client side actions that load data. I never imagine sending anyone to the actual raw url we're discussing here with the path in it.

indirect:
Backing all the way up, I'd like to reiterate the extremely high level set of goals we (eventually) want to work towards, built on top of the ingestion system:

1. content-addressed files API
2. gem contents API
  - uses the content-addressed API to return the file with eg hash abc123 when eg rake-1.0.0/lib/rake.rb is requested
  - is probably the basis for the future code search
3. fancy gem version content HTML pages (generated by rails app, cached by fastly)
  - renders a gem version contents as fancy pages a la github’s repo browser
4. fancy gem version diff HTML pages (generated by rails app, cached by fastly)
  - renders diffs between gem version contents as fancy pages a la github’s diff viewer
5. fancy code search result HTML pages (generated by rails app from sourcegraph results? something else?)
  - uses a search cluster to render results from one version, all versions of a named gem, or all versions of all gems
  - let's see if we can use sourcegraph.com for this
6. (maybe?) fancy code browsing, search, and navigation a la GitHub. maybe we just programmatically create repos for every gem and push new commits with the contents of every version. lol.

Moving back from those extremely high level goals to acceptance criteria for just this particular step, I think the access patterns this file and db schema will need to support are:

1. easily read a single file from a gem version (eg, `gem contents rack 1.0.0 lib/rack.rb` or `GET api/gems/rack/1.0.0/files/lib/rack.rb` can print the file contents, with just one HTTP request)
2. provide a manifest with file checksums for the entire gem, for possible users or future tools to use, something like `GET /api/gems/rack/1.0.0/manifest`. The manifest should include (at least) the file name and the content hash with the hashing algorithm included, like `sha256-abc123`.
3. provide file content when requested by hash, something like `GET /api/gems/rack/1.0.0/content/sha256-abc123`.

I'm not 100% sure about number 3. I might be able to be talked out of it.

Within those constraints, I think any solution that provides the API above and scales to about one pushed gem every 2-5 seconds is a workable solution.

jchestershopify:
> Backing all the way up, I'd like to reiterate the extremely high level set of goals we (eventually) want to work towards, built on top of the ingestion system:
> content-addressed files API

I think I'd like to pop the stack one more time. I get that the idea is to allow folks to retrieve gem contents without downloading and extracting the gem. What I am unclear about is for whom this is make-or-break functionality. Who is going to use this functionality? How many? How essential to their workflow is it in order to justify RubyGems handling the additional complexity and expense? Right now this effort seems to be justified in terms of itself, rather than referring to an external need.

text/0008-ingest-gem-content.md
martinemde: SRI, at least browser implementations, would seem to encourage sha512 and the format requires base64 encoding. Base64 isn't directly url safe, so using it for urls may not work as smoothly as hex encoding. I like the idea of adding this integrity attribute for each file, but I'm not sure the browser supported format would work for our URLs. We could deviate from the browser format or use a second sha256 that just stands in for the key.

I'm not worried about integrity as much for deduplicating file keys. It's infinitesimally unlikely to have 2 valid textual files in a gem that share a sha256.

indirect: hex-encoded sha256 seems fine as a starting point. the chance that we will eventually want to migrate to a different hash function for security reasons are basically 100%, so I think we should build it in a way that allows us to migrate hash versions in the future. maybe that just means (in the data model) a column for algorithm and it's an enum with only one value so far.

indirect:
Good point! We can start there. 😄 

Problem statement: RubyGems.org hosts many packages that our users are going to install on their machines and then execute. It's important for users to be able to see 1) what code you are about to run, and 2) view diffs between versions when upgrading.

We can't just refer users to GitHub diff links because 1) gems are not required to provide their source in any publicly available repository, and 2) the `gem build` process can modify source files arbitrarily before tarring up a .gem, so even an official git repo is not the same as what is inside the gem.

The "diff between versions" feature is useful and (imo) necessary for RubyGems.org to provide. Since we definitely want to build that feature, it makes sense to also lay the groundwork so that we can potentially display authoritative contents of packages, for users to view, or for other tools that might be built.

jchestershopify:
OK, I see the underlying cause - thankyou for the clear explanation @indirect! Could we expand the RFC to take this in? It may help a future passer-by to get their bearings.

text/0008-ingest-gem-content.md
martinemde: I'd like to settle on some algorithm for naming files, and then add an additional column for integrity hashes later if we deem the current algorithm to be insufficient. It's one thing to add a new column and another thing entirely to rewrite the filename of every file already stored, or otherwise give up deduplication at the time when we switch.

martinemde:
@jchestershopify @indirect I updated the `## Motivation` section based on the feedback and discussion. Hopefully that captures it better. Thank you for helping us improve this proposal!

simi:
To be honest, even after reading `Motivation` part, as a RubyGems user, I don't really see the value in building this. I had no need for this in all my "Ruby life". To see changes today, you can use "Review changes" link on gems page (https://rubygems.org/gems/rubygems-update pointing to https://my.diffend.io/gems/rubygems-update/prev/3.4.5). It is using `diffend.io` and there is no need to maintain this from limited power of 'rubygems.org` resources. To explore gem content, it is super simple to download, unpack and locally explore with my common tools like favorite editor.

The interesting part could be for maintainers to be able to cross-search across gems and versions. Something similar could be achieved today with [gem-codesearch](https://github.com/akr/gem-codesearch). It is actively used by @hsbt to check on potential breaking changes (like https://github.com/rubygems/rubygems/pull/6311#issuecomment-1404556676). But I'm not sure it is possible to build anything fast enough to be able to respond to HTTP request in reasonable time for this purpose with the current amount of gem codes. @hsbt would you mind to share some stats from gem-mirror size these days? Looking into sum of all versions from production DB it seems all gems together are around 728 GB today (packed). We can also ask other package managers (like NPM - https://www.npmjs.com/package/@rails/webpacker?activeTab=explore) about their experience and implementation details on this.

I would like to suggest to first decide what's the main reason for this functionality (user vs maintainer).

For user's one, it would be great to start with something simple (if we decide to move this forward). `rubydoc.info` like lazy approach seems nice to me. Build side-app downloading and unpacking gems on request somewhere to temporary storage with simple code browser (linked from RubyGems.org gem show page).

Other alternative would be to provide simple way (link on RubyGems page) of onboarding gem into any of the external online code editors (like [GitHub Codespaces](https://github.com/features/codespaces), [CodePen](https://codepen.io/), [CodeSandbox](https://codesandbox.io/, [StackBlitz](https://stackblitz.com/), ...).

hsbt:
> @hsbt would you mind to share some stats from gem-mirror size these days?

It's 201 GB.

In my view, I'm enough to investigate gem content by gem-codesearch and contents provide by dependebot alert like https://github.com/ruby/rubyci/pull/362 and https://github.com/puma/puma/compare/v6.0.1...v6.0.2 .

>I would like to suggest to first decide what's the main reason for this functionality (user vs maintainer).

👍 

IMO, this proposal is good rather than none if we have unlimited network, storage and CPU. I don't feel like the benefits are worth the cost now.

indirect:
Thank you for the feedback. 🙏🏻

It is true that we can use Dependabot, GitHub, Diffend, and gem-mirror to read, navigate, and search gems today. Combined, these are a perfectly okay way to check on gems.

One of the tasks I am working on this year is how to improve RubyGems.org for both maintainers and for end-users. There are many items on both lists. One item for maintainers is gem code search. One item for end-users is viewing gem diffs between versions, as well as READMEs, CHANGELOGs, and other files inside a gem version.

I think it is worthwhile to build these tools into RubyGems.org itself. RubyGems is 20 years old this year, even older than git (!), and I want RubyGems to be able to continue for 20 more years. I believe that means RubyGems should be able to function without needing GitHub or Diffend to exist.

Right now we have enough funding (thank you Shopify and Sovereign Tech Fund) that this is affordable, and we can even work on other improvements at the same time. I am happy to discuss prioritizing improvements with all of you, but let's do that somewhere else so this poor RFC can rest. 😄 

In summary:
- gem contents and version diffs are mainly for ruby developer "users"
- code search is mainly for gem developers, rubygems/bundler maintainers, and ruby-core "users"
- this RFC can provide a base for gem contents, version diffs, and code search to build on top
- we can build gem contents, version diffs, and code search while also working on other important improvements
- I think we should do this so that we do not depend on any one company for this useful functionality

martinemde:
I've updated the RFC to capture the current behavior in rubygems/rubygems.org#3454.

indirect:
Merging this RFC now that the implementation has been merged. 👍🏻

PR #45, opened 2023-01-25
Add RFC for gem exec command

segiddins:


indirect (approved):
Solid. I will use this once we have it.

text/0000-gem-exec.md
```diff
@@ -0,0 +1,66 @@
...
+# Guide-level explanation
+
+The `gem exec` command can be thought of as a shortcut to running `gem install` and then the executable from the installed gem.
+
+For example, `gem exec rails new .` will run `rails new .` in the current directory, without having to manually run `gem install rails` only to run `rails new` once, and no other `rails` commands (because `rails new` generates a Gemfile).
+
+For a more complete example:
+
+```sh
+$ gem exec --gem cocoapods -r '> 1' -r '< 1.3' -v -- pod install --no-color --help
```
simi, line 23: It would be great to support also version identifier like you can do with `gem install`. `gem exec rails:7.1 new .`

simi:
:thinking: `gem exec rails new .` is real selling point on this. :+1:

text/0000-gem-exec.md
segiddins: that makes sense too, I'll add support for it.

hsbt (approved):
👍

text/0000-gem-exec.md
```diff
@@ -0,0 +1,95 @@
...
+If the binary is not found, the command will install the gem (by default of the same name as the binary, though it can be set by a command line option),
+and then activate and execute that binary.
+
+# Drawbacks
+
+This proposal should have no impact on existing users, since it is purely additive.
+
+One potential area of confusion is with Bundler / users using Gemfiles.
+To start, we plan on ignoring the Gemfile entirely when running `gem exec`, which may be confusing if the user is attempting to use the Gemfile to customize gem sources.
+Additionally, `gem exec` is confusable with the `bundle exec` command, and they will work differently. A potential way to ameliorate that confusion is for `gem exec` to check if the requested executable exists in the current bundle, and advise the user to use `bundle exec` in that case.
```
simi, line 83: Maybe we can find better alternative name other than `gem exec` to prevent `bundle exec` confusion. Maybe `gem run`?

text/0000-gem-exec.md
```diff
@@ -0,0 +1,95 @@
...
+
+# Rationale and Alternatives
+
+We could continue to point users to the existing [`gemx`](https://github.com/rubygems/gemx) gem to solve this problem, or encourage using `gem install $gem && $gemBinary`, but those solutions (which have existed for years) are not sufficiently discoverable.
+
+Implementing this proposal will make it easy for users to run one-off commands that _happen_ to be shipped as gems, and instructions to `gem exec $gem` can become widespread.
+
+# Unresolved questions
+
+Should there be an option to use an isolated gem path for `gem exec` installs? Should it be required or the default? Should we look into using existing gems but not installing into the default path?
```
simi, line 93: > Should we look into using existing gems

What is actual plan for this? Let's say there is locally installed `rails 7.0.0` and `gem exec rails -r "~> 7.0` is requested. There is also newer version available remotely. Is it going to first install latest version matching related constraint and run `rails` command in there or is it going to be satisfied with local already installed matching version?

text/0000-gem-exec.md
segiddins: hmm, is it worth diverging from `npm exec` over this?

text/0000-gem-exec.md
segiddins: current plan is to not use existing gem unless `--conservative` is given

indirect: I think it makes sense for the default usecase to be installing and then running the latest version of a package, and only skipping the install if you already have the newest version installed. If you want to run whatever you already have installed without installing the latest, you can probably run `rails new .` instead of `gem exec rails new .`, right?

indirect: Given that we default to not using the already-installed gem, I think it's reasonable to have a separate path for installs without the conservative flag, so you can `gem exec` without filling up your global gemset.

indirect (approved):

simi:
:partying_face:

PR #46, opened 2023-01-27
Add RFC to deprecate the dependency API on rubygems.org

segiddins:
[Rendered](https://github.com/segiddins/rfcs/blob/segiddins/deprecate-dependency-api/0000-deprecate-dependency-api.md)

### DIFF
--- a/0000-deprecate-dependency-api.md
+++ b/0000-deprecate-dependency-api.md
@@ -1,0 +1,40 @@
+- Feature Name: Deprecating the RubyGems.org dependency API
+- Start Date: 2023-01-27
+- RFC PR: (leave this empty)
+- Bundler Issue: (leave this empty)
+
+# Summary
+
+Plan for the deprecation and removal of the "dependency API" (`/api/v1/dependencies`) on rubygems.org.
+
+# Motivation
+
+The dependency API was the primary way Bundler fetched dependency info for Gemfile resolution until the release of [Bundler 1.12](https://bundler.io/blog/2016/04/28/the-new-index-format-fastly-and-bundler-1-12.html) in April 2016, which introduced the ["new" Compact Index API](https://andre.arko.net/2014/03/28/the-new-rubygems-index-format/).
+
+The Compact Index was designed to be completely cacheable and served at the edge by Fastly. In contrast, the dependency API is completely _uncacheable_, and requires RubyGems.org to run a complex database query for each request, accounting for almost 50% of the total uncached request volume to RubyGems.org with nearly 3000qps.
+
+Because the Compact Index has been out for over half a decade and Bundler & RubyGems both support fallback to the "full" index (`specs.4.8.gz` files) when the Dependency API is unavailable, we plan to deprecate and delete the dependency API over the coming months.
+
+# Guide-level explanation
+
+This will be a multi-phase transition.
+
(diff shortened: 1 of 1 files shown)


indirect:
💯, no notes. Let's do this.

indirect (approved):

simi (approved):

simi:
Btw. I have related plan to document RubyGems Gemserver API (only the part needed for gem client and bundler) and prepare a API linter so other gemservers can easily follow. This linter can also inform about those deprecations (or other upcoming changes) to make ti simple to stay in sync.

Those are my initial notes:

```
create simple RubyGems API specification and linter (to make it clear which endpoints needs to be implemented to implement valid minimal RubyGems API source)
- use for RubyGems.org as part of a test suite
- promote to most common OSS private sources (https://github.com/rubygems/gemstash,h ttps://github.com/geminabox/geminabox)
- try to promote to private hosters (GitHub Packages, ...)
```

Maybe we can sync and release this documentation or linter (or even both) at the time of announcing this deprecation as well.

simi (approved):

0000-deprecate-dependency-api.md
```diff
@@ -0,0 +1,40 @@
...
+
+Plan for the deprecation and removal of the "dependency API" (`/api/v1/dependencies`) on rubygems.org.
+
+# Motivation
+
+The dependency API was the primary way Bundler fetched dependency info for Gemfile resolution until the release of [Bundler 1.12](https://bundler.io/blog/2016/04/28/the-new-index-format-fastly-and-bundler-1-12.html) in April 2016, which introduced the ["new" Compact Index API](https://andre.arko.net/2014/03/28/the-new-rubygems-index-format/).
+
+The Compact Index was designed to be completely cacheable and served at the edge by Fastly. In contrast, the dependency API is completely _uncacheable_, and requires RubyGems.org to run a complex database query for each request, accounting for almost 50% of the total uncached request volume to RubyGems.org with nearly 3000qps.
+
+Because the Compact Index has been out for over half a decade and Bundler & RubyGems both support fallback to the "full" index (`specs.4.8.gz` files) when the Dependency API is unavailable, we plan to deprecate and delete the dependency API over the coming months.
```
simi, line 16: 🤔 Is there any known gem server (public or private) supporting Dependency API (in production graded state) as well?

segiddins: I think gemstash and artifactory both support the dependency API, but artifactory also (as of the latest version) supports the compact index

indirect: I think it's also OK to leave in client support for the dependency API (rubygems or bundler or both) so that existing servers can continue to provide that benefit if they do not yet support the compact index. The main goal here is to remove support from rubygems.org so that we can remove a bunch of deprecated code paths. It will make Bundler versions released between 2011 and 2016 somewhat slower to install, but I think that is OK now. We should ask anyone who complains to upgrade to Bundler newer than 7 years old. :)

segiddins:
https://github.com/rubygems/rubygems.github.io/pull/145 is the matching blog post, let's ship this?

0000-deprecate-dependency-api.md
simi: Yes, I was about to make it suuuuuuuuuper clear it is not going to be removed from clients for now.

PR #48, opened 2023-03-27
Add RFC for time series gem downloads

segiddins:
@indirect would appreciate your help rounding this off

### DIFF
--- a/text/0000-time-series-gem-downloads.md
+++ b/text/0000-time-series-gem-downloads.md
@@ -1,0 +1,199 @@
+- Feature Name: Store gem download counts in a timeseries database
+- Start Date: 2023-03-16
+- RFC PR: (leave this empty)
+- Bundler Issue: (leave this empty)
+
+# Summary
+
+We will move to storing download counts in a timeseries database (specifically, timescale), to enable displaying download counts (for all gems, all versions of a single gem, or a single version) grouped by time period (e.g. the last day, month, or year).
+
+# Motivation
+
+Users would like to see this data.
+We would like to make insertion of download counts require no contention on the production postgres database.
+We would like to avoid maintaining bespoke infrastructure for download count tracking.
+We want to migrate to the new storage system without losing historical download counts.
+
+As a followup, we will work on displaying the new data in RubyGems.org, but that design work is out of scope for this RFC.
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Guide-level explanation
+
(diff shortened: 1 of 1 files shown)


indirect:
@segiddins this looks good to me, I think. my only remaining question is how we will ensure the main app doesn't break if e.g. timescale disappears. for the work outlined in this RFC, I think the only thing that might break is the background job that writes to timescale? if so, that means it's mainly a question for the future work that will display download counts and graphs.

segiddins:
So my thought there is, we can have a cron job that copies all time download numbers to the main postgres database on some cadence?

indirect:
Oh yeah, that seems good. That plus a circuit-breaker type thing on the requests to the timescaledb so puma can continue to serve other rails requests sounds like it should work.

segiddins:
> That plus a circuit-breaker type thing on the requests to the timescaledb so puma can continue to serve other rails requests sounds like it should work.

I'm not quite sure what you mean by that? My thinking was we'd use stimulus to show download counts on pages, so as to be able to cache download numbers separately, and that would give us a natural place to keep reads from timescale off the critical path

indirect:
That's pretty much what I was thinking--something out of the critical path to show the numbers, something out of the critical path to write into timescale. Thanks for writing this up!

text/0000-time-series-gem-downloads.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+Note that the raw `Download` class will not be exposed anywhere -- as far as the rails app is concerned, it's a write-only table, and all queries will happen against the continuous aggregates.
+
+# Drawbacks
+
+- Adding another datastore complicates contributing to the application
+- Another datastore means we can't natively write joins against the download data
+- Timescale cloud costs money
+- There is no terraform provider for timescale cloud (yet) requiring us to write one or manually provision
```
jonatas, line 189: Now, there's the official https://github.com/timescale/terraform-provider-timescale  ✅

text/0000-time-series-gem-downloads.md
```diff
@@ -0,0 +1,205 @@
...
+- This is still high cardinality data, and there's no knowing for sure how well timescale will compress it
```
jonatas, line 190: I can help testing and benchmarking it. Is there any data available to seed the database? Or if anyone can share any numbers to give me some insights about how big is the cardinality.

indirect (approved):

PR #50, opened 2023-08-29
RFC for bundler checksum verification

martinemde:
This is a draft based on the feature work in rubygems/rubygems#6374.

### DIFF
--- a/text/0011-gem-checksum-verification.md
+++ b/text/0011-gem-checksum-verification.md
@@ -9,19 +9,20 @@ Bundler is adding checksum verification of gems when they are installed. It shou
 
 # Motivation
 
-Verifying gem checksums with known checksums at install time is a stronger way to verify that the exact same gem source is being used in every environment where the bundle is installed. Using checksums, especially those sourced from rubygems.org, is a more strict way of locking to a specific gem version. The feature should work transparently in much the same way that bundler already locks to specific versions.
+Verifying gem checksums with known checksums at install time is a stronger way to verify that the exact same gem source is being used in every environment where the bundle is installed.
+Using checksums, sourced from rubygems.org or from digests of local .gem file, is a more strict way of locking to a specific gem version.
+The feature should work transparently in much the same way that bundler already locks to specific versions by ensuring that not just the same version, but the exact same rubygem file data, is installed on each environment.
 
 # Guide-level explanation
 
-Upon first upgrading to the version of Bundler with this feature, the checksum feature will be enabled.
+Upon first upgrading to this version of Bundler, expect bundler to continue to be secure by default and to progressively add checksums to the lockfile.
 
-First usage will fall into two use cases:
-1. Users that want to immediately use the new checksums feature.
-2. Users that are not specifically using the checksums feature, but expect bundler to work securely without interference.
+If you wish to immediately add checksums to your lockfile for all locked gems, run `bundle lock`.
+Bundle lock now fetches checksums from remote sources by default.
+If you would like to bypass this behavior, add the `--no-checksums` flag.
 
-For users excited about the new feature, the usual commands and docs should show the new options for securing the bundle with checksums. Documentation and the release notice should include instructions for directly using the new feature.
-
-For users that are not specifically looking for the feature, common commands like `bundle install`, `bundle update`, `bundle lock`, `bundle add`, should automatically make use of checksums when available. When there are missing checksums that must be fetched with a specific interaction, the UI should say what that action is.
+Common Bundler commands like `bundle install`, `bundle update`, `bundle lock`, `bundle add` will now automatically record and verify checksums.
+Commands that rely on checksums for verification will output a message or warning when checksums are missing or mismatched.
 
 Example:
 
@@ -29,13 +30,16 @@ Example:
 $ bundle install
 Bundle complete! 88 Gemfile dependencies, 256 gems now installed.
 Use `bundle info [gemname]` to see where a bundled gem is installed.
-256 gems are missing checksums. Source integrity could not be verified.
-Use `bundle lock` to add checksums to Gemfile.lock.
+Use `bundle lock` to add missing checksums to Gemfile.lock.
 ```
 
-When a gem version is updated in the lockfile using a source that has checksums (like rubygems.org), the checksum should be recorded. Running `bundle update` or `bundle add` should record the checksum from the source into the Gemfile.lock. If a checksum is not available from the source because the source does not provide such info, then a checksum should be created during install using the .gem file or a warning should be printed.
+Running `bundle update` or `bundle add` will record the checksum from the source (e.g. rubygems.org) into the Gemfile.lock, if it is available.
+If a checksum is not available from the source because the source does not provide such info (e.g. private gemservers) then a checksum will be created during install using the .gem file.
+If a checksum can't be created because the source is a path or git source, then only the gems name and version will be recorded with a blank checksum.
 
-If a user wishes to ensure that the environment they are using contains only gems installed from verified sources that match the checksums in the Gemfile.lock, the command `bundle pristine` should be run. The `bundle pristine` command installs every gem fresh from the .gem source, but does not fetch gems from the server or read the source index. Therefore, if a users wants all gems to be verified upon install, `bundle pristine` should be run after checksums are written to the Gemfile.lock.
+If you want to ensure that the bundled environment contains only gems matching the checksums in the lockfile, run `bundle pristine`.
+The `bundle pristine` command installs every gem fresh from a newly downloaded .gem source.
+The pristine install will trigger computation and comparison of the generated SHA256 checksum with the checksum stored in the lockfile.
 
 Example:
 
@@ -44,82 +48,72 @@ $ bundle pristine
 Installing rake 13.0.6
 …
 Installing rspec 3.12.0
-42 gems are missing checksums. Source integrity could not be verified.
+42 gems without checksums.
 Use `bundle lock` to add checksums to Gemfile.lock.
-$ bundle lock --update
+$ bundle lock
 Writing lockfile to path/to/Gemfile.lock
 $ bundle pristine
-Installing rake 13.0.6 (verified)
+Installing rake 13.0.6
 …
-Installing rspec 3.12.0 (verified)
+Installing rspec 3.12.0
 $
 ```
 
-If a gem that is about to be installed does not match the checksum, an error is generated.
+When bundler installs a gem from a `.gem` file, it computes the SHA256 checksum of the file.
+If an existing checksum is available from the lockfile or the remote source, it will be compared with the computed checksum at install time.
+If the checksums do not match, an error is generated and installation is halted.
+If no checksum is recorded in the lockfile, the computed checksum is saved to the lockfile where future installs can verify that the same gem is installed.
 
 Example:
 
 ```
 $ bundle install
 Installing rake 13.0.6
-Bundler cannot continue installing rake-13.0.6.
-The checksum for the downloaded `rake-13.0.6.gem` does not match the known checksum for the gem. 
+Bundler found mismatched checksums. This is a potential security risk.
+  rake (13.0.6) sha256-2222222222222222222222222222222222222222222222222222222222222222
+    form the lockfile CHECKSUMS at Gemfile.lock:21:17
+  rake (13.0.6) sha256-814828c34f1315d7e7b7e8295184577cc4e969bad6156ac069d02d63f58d82e8
+    from the gem at path/to/rake-13.0.6.gem
 
-Expected: sha256-814828c34f1315d7e7b7e8295184577cc4e969bad6156ac069d02d63f58d82e8 from: Gemfile.lock:21:1 CHECKSUMS rake (13.0.6)
-Downloaded: sha256-69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b6 from:
-path/to/rake-13.0.6.gem SHA256 hexdigest
+  To resolve this issue you can either:
+    1. remove the gem at path/to/rake-13.0.6.gem
+    2. run `bundle install`
+  or if you are sure that the new checksum from the lockfile CHECKSUMS at Gemfile.lock:21:17 is correct:
+    1. remove the matching checksum in Gemfile.lock:21:17
+    2. run `bundle install`
 
-To resolve this issue:
-Delete the downloaded gem files referenced above and run `bundle install`
-
-If you are sure that the downloaded gem is correct, you can run…
-(RFC TODO: Currently undecided on what the command should be here to allow unverified gems.)
-`bundle install --skip-verification` OR
-`bundle config set unverified ignore` OR
-Remove the offending line from CHECKSUMS in Gemfile.lock
+To ignore checksum security warnings, disable checksum validation with
+  `bundle config set --local disable_checksum_validation true`
 ```
 
-Certain checksums will always be unavailable because the source does not provide a checksum. When a checksum is not available, a line can be added to the Gemfile.lock CHECKSUMS section that explicitly indicates that no checksum is recorded for that gem version. Future installs of gems with intentionally ignored checksums should not error or warn when the gem is installed. The **(RFC TODO)** command will allow idicating that a gem version should ignore checksum verifications. Alternatively, the SHA256 digest of the gem at install time can be recorded allowing the user to verify future installs do not deviate from the recorded checksum.
+Certain checksums will always be unavailable because the source does not provide a checksum.
+When a checksum is not available, the gem will be added to the Gemfile.lock CHECKSUMS section without a checksum.
 
-Platforms may pose a problem for Gemfile.lock checksums. If a user has not added one of the platforms where they bundle their app, then checksum verification could be skipped. In order to avoid this security vulnerability, the following should happen automatically.
-
-1. When a gem is installed and no checksum is available in the lockfile, a warning should be printed indicating that unverified gems were installed.
-2. When a gem is installed where a checksum is available for the same gem on a different platform, an error should be raised. (i.e. if nokogiri for darwin has a checksum, but nokogiri for musl does not, then an error should be raised).
-3. When bundle install is configured with frozen, an error should be raised for any gem installed that does not have a checksum unless the checksum is explicitly ignored.
-
-When platform or missing checksum issues are raised, the error should always indicate what command or process should be performed to fix the error.
-
-Bundling the app on another platform, in CI for example, should highlight any platform issues. This will make the bundler upgrade easier for users.
-
-To encourage uptake of this security feature, it should be enabled by default. Any actions that fetch authoritative checksums should silently record them in the Gemfile.lock. Progressive upgrading to verified checksum behavior should happen naturally unless explicitly disabled.
-
-Gem upgrades in the Gemfile.lock happen in two ways currently. Each way will behave differently.
-
-1. The user explicitly runs `bundle update`.
-2. Dependabot, or another automation, changes the version of a locked gem directly in the Gemfile.lock (assuming current Dependabot behavior without a corresponding upgrade to match this feature).
-
-Explicit gem updates will record the new checksum for the gem and remove any old checksums for gems that are no longer in the Gemfile.lock.
-
-Direct updates to the Gemfile.lock will print a warning during install. The warning should say that X number of gem(s) could not be verified.
-If the bundle is frozen, then the warning should instead print an error and halt gem installation.
+Users should be aware that when bundling on CI or production, new gems can be added for a platform not found in the Gemfile.lock.
+Bundler will silently record the new checksums for the missing gem just like on a local development machine.
+If you would like to ensure that only lockfile checksums are used, bundle install should use use the frozen or deployment configurations.
 
 # Reference-level explanation
 
-Gem checksums are fetched from rubygems.org as part of the compact index. The checksums are stored in memory during the bundle process.
+Gem checksums are fetched from rubygems.org as part of the compact index.
+The checksums are stored in memory during the bundle process.
 
 If any gems are recorded from a separate source or installed via .gem file, another checksum is recorded and compared with the original.
 
-If any of the checksums for the same gem name, version and platform differ, an error is raised.
+If any of the checksums for the same gem name, version and platform differ, an error is raised with instructions for resolving the error.
 
 When the Gemfile.lock is written, a new CHECKSUMS section is written with all the gems in the bundle and their corresponding checksums.
 
-Future compatibility for new checksum algorithms is supported by reading and writing existing checksums for installed gems even if the algorithm is unknown. Checksum comparisons take into account the algorithm used and raises errors accordingly.
+Future compatibility for new checksum algorithms is supported by reading and writing existing checksums for installed gems even if the algorithm is unknown.
+Checksum comparisons take into account the algorithm used and raises errors accordingly.
 
 Bundler commands that interact with checksums either fetch checksums from sources and update the Gemfile.lock or compare checksums in the Gemfile.lock with the computed digest of the gem being installed.
 
-Internal storage of checksums indexes the checksums by a gem's full name (name-version-platform) and the checksum's algorithm. The source of each checksum is stored with the checksum so that errors can detail where conflicting checksums came from.
+Internal storage of checksums indexes the checksums by a gem's NameTuple (name, version, platform) and the checksum's algorithm.
+The source of each checksum is stored with the checksum so that errors can describe how to fix conflicting checksums.
 
-When two sources have a checksum for the same gem full name, the are compared. If they are the same, they are merged into a single record that associates the checksum with all of the sources for that checksum.
+When two remote sources have a checksum for the same gem, the are compared.
+If they are the same, bundler proceeds as normal
 
 ### Example CHECKSUM section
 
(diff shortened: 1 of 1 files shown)


text/0011-gem-checksum-verification.md
```diff
@@ -150,3 +179,7 @@ If a user configures CI to use locked gem versions, it would be necessary for De
 
 - What are the exact commands, options, warning and error messages that are used to interact with the feature?
 - What are the unexpected errors that may happen the first time people verify the gems they are downloading? Are there any errors or problems in the existing rubygems.org repository?
+
+### How do we handle confusion about the authority of checksums written to the Gemfile.lock
+
+The source of checksums in the Gemfile.lock becomes a matter of trust once it's written. Did the checksum come from the API or was it calculated from a .gem file on a developers computer. If a checksum error is resolved by one developer in a way that saves an incorrect checksum, how should people know when to approve these changes or not. It may not even be common practice for most teams to look at the Gemfile.lock, and changes can often be hidden in pull request reviews. Without a process for checking that the checksums are trustworthy, it's left to every development team to decide on a process. One solution would be a bundle command that could be run in CI every time the gems are installed that verifies the authenticity of checksums in the Gemfile.lock.
```
olleolleolle, line 184: I found there were sentences that could in a question mark:

```suggestion
The source of checksums in the Gemfile.lock becomes a matter of trust once it's written. Did the checksum come from the API or was it calculated from a .gem file on a developers computer? If a checksum error is resolved by one developer in a way that saves an incorrect checksum, how should people know when to approve these changes or not? It may not even be common practice for most teams to look at the Gemfile.lock, and changes can often be hidden in pull request reviews. Without a process for checking that the checksums are trustworthy, it's left to every development team to decide on a process. One solution would be a bundle command that could be run in CI every time the gems are installed that verifies the authenticity of checksums in the Gemfile.lock.
```

olleolleolle (commented):
Typography comment.

text/0011-gem-checksum-verification.md
martinemde: I appreciate you taking the tike to read the RFC. Thank you for the fix!

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,184 @@
...
+Delete the downloaded gem files referenced above and run `bundle install`
+
+If you are sure that the downloaded gem is correct, you can run…
+(RFC TODO: Currently undecided on what the command should be here to allow unverified gems.)
+`bundle install --skip-verification` OR
+`bundle config set unverified ignore` OR
+Remove the offending line from CHECKSUMS in Gemfile.lock
+```
+
+Certain checksums will always be unavailable because the source does not provide a checksum. When a checksum is not available, a line can be added to the Gemfile.lock CHECKSUMS section that explicitly indicates that no checksum is recorded for that gem version. Future installs of gems with intentionally ignored checksums should not error or warn when the gem is installed. The **(RFC TODO)** command will allow idicating that a gem version should ignore checksum verifications. Alternatively, the SHA256 digest of the gem at install time can be recorded allowing the user to verify future installs do not deviate from the recorded checksum.
```
segiddins, line 82: TODO here

segiddins, line 82: > Alternatively, the SHA256 digest of the gem at install time can be recorded allowing the user to verify future installs do not deviate from the recorded checksum.

I like this, it matches with the `lock on first install` approach bundler already takes with dependency resolution

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,184 @@
...
+
+Platforms may pose a problem for Gemfile.lock checksums. If a user has not added one of the platforms where they bundle their app, then checksum verification could be skipped. In order to avoid this security vulnerability, the following should happen automatically.
+
+1. When a gem is installed and no checksum is available in the lockfile, a warning should be printed indicating that unverified gems were installed.
```
segiddins, line 86: should this be printed even if there's a checksum coming from the remote source? if so, that feels like it would get annoying

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,184 @@
...
+2. When a gem is installed where a checksum is available for the same gem on a different platform, an error should be raised. (i.e. if nokogiri for darwin has a checksum, but nokogiri for musl does not, then an error should be raised).
```
segiddins, line 87: this will be a breaking change to how bundler handles formats, and I would anticipate user backlash against this. (doubly so if, as above, it happens even when a checksum comes from the remote). Currently, you can transparent add a platform to the lock file via `bundle install`, and it is only disallowed under `--deployment` and `--frozen`

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,184 @@
...
+
+# Unresolved questions
+
+- What are the exact commands, options, warning and error messages that are used to interact with the feature?
+- What are the unexpected errors that may happen the first time people verify the gems they are downloading? Are there any errors or problems in the existing rubygems.org repository?
+
+### How do we handle confusion about the authority of checksums written to the Gemfile.lock
+
+The source of checksums in the Gemfile.lock becomes a matter of trust once it's written. Did the checksum come from the API or was it calculated from a .gem file on a developers computer? If a checksum error is resolved by one developer in a way that saves an incorrect checksum, how should people know when to approve these changes or not? It may not even be common practice for most teams to look at the Gemfile.lock, and changes can often be hidden in pull request reviews. Without a process for checking that the checksums are trustworthy, it's left to every development team to decide on a process. One solution would be a bundle command that could be run in CI every time the gems are installed that verifies the authenticity of checksums in the Gemfile.lock.
```
segiddins, line 184: I would look to other package managers here and see what they do (which in my experience is usually strictly additive?)

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,184 @@
...
+
+Expected: sha256-814828c34f1315d7e7b7e8295184577cc4e969bad6156ac069d02d63f58d82e8 from: Gemfile.lock:21:1 CHECKSUMS rake (13.0.6)
+Downloaded: sha256-69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b69b6 from:
+path/to/rake-13.0.6.gem SHA256 hexdigest
+
+To resolve this issue:
+Delete the downloaded gem files referenced above and run `bundle install`
+
+If you are sure that the downloaded gem is correct, you can run…
+(RFC TODO: Currently undecided on what the command should be here to allow unverified gems.)
```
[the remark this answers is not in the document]
indirect: Yeah, I think `gem "foo", checksum_override: "abc123"` is probably the most reasonable way to do this... if we actually want to allow it? I'm not sure if we do.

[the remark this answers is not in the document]
segiddins: it would have to be a hash, because each platform would have its own checksum.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,184 @@
...
+
+```
+$ bundle pristine
+Installing rake 13.0.6
+…
+Installing rspec 3.12.0
+42 gems are missing checksums. Source integrity could not be verified.
+Use `bundle lock` to add checksums to Gemfile.lock.
+$ bundle lock --update
+Writing lockfile to path/to/Gemfile.lock
```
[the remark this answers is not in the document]
martinemde: I agree. I think I added this as an idea but have since reconsidered and I don't plan to implement it. I will update.

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
martinemde: The error could output the command to run to ignore this specific gem's mismatch and then we would just record it in the gemfile.lock. I imagine just changing the relevant line in the CHECKSUMS section to `rake (1.0.0) IGNORED`. Then since it's transiently in the lockfile, the next update to a different version would either produce a new error or resolve correctly.

text/0011-gem-checksum-verification.md
martinemde: Great. This is how it is implemented right now for .gem files.

There's still git or path gems that will never have checksums. I don't think we need any explicit ignore for them, as I started with at the top. I can update that.

text/0011-gem-checksum-verification.md
martinemde: Yes, there's a conflict between storing the checksums from the .gem on install and not installing unverified gems in production when the platform is missing. I think as you commented below, we want to be transparent about this unless the bundle is frozen.

I think we need to add this feature then: If new checksums are recorded during install while the bundle is frozen, then fail.

text/0011-gem-checksum-verification.md
martinemde: I agree. I'll remove this line. I think I was imagining a frozen gemfile when I wrote this, but then I wrote the next like 3. which deals with frozen gemfiles.

text/0011-gem-checksum-verification.md
martinemde: Hmm, let me refine what I said there, because I don't think it captures the whole picture and I think there are conflicts.

Let's just talk about the frozen bundle situation.

Here's the bundler docs about frozen:

> --frozen
> Do not allow the Gemfile.lock to be updated after this install. Exits non-zero if there are going to be changes to the Gemfile.lock.
> This option is deprecated in favor of the frozen setting.

Given the lock:
```
CHECKSUMS
  foo (1.0.0)
  bar (1.0.0-darwin-23)
```

When does installing the gem fail during a frozen bundle?

1. Install foo. A checksum is computed at install, but not recorded because the bundle is frozen.
2. Install bar (1.0.0-x86_64-linux) which is entirely missing from the CHECKSUMS section.
3. Install baz which is entirely missing from the CHECKSUMS section.
4. All of the above because if you're running in frozen mode, the docs say it will fail if any update would produce a change to the lock.

This is a real challenge because this is going to happen to a lot of people:
1. Update bundler version
2. Run bundle install (no checksums are written unless the gems are not already installed)
3. Commit and push to production
4. Frozen install fails because new checksums are found.

This means we need to make exceptions for checksums being found from gem installs even when frozen. Or we need to say "if you're running in frozen mode, all checksums must be recorded that can be recorded". This means running `bundle lock` before you commit and push and might require that we inform people when we first write the CHECKSUMS section when it didn't exist before.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,184 @@
...
+
+```
+$ bundle install
+Bundle complete! 88 Gemfile dependencies, 256 gems now installed.
+Use `bundle info [gemname]` to see where a bundled gem is installed.
+256 gems are missing checksums. Source integrity could not be verified.
+Use `bundle lock` to add checksums to Gemfile.lock.
+```
+
+When a gem version is updated in the lockfile using a source that has checksums (like rubygems.org), the checksum should be recorded. Running `bundle update` or `bundle add` should record the checksum from the source into the Gemfile.lock. If a checksum is not available from the source because the source does not provide such info, then a checksum should be created during install using the .gem file or a warning should be printed.
```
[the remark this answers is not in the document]
martinemde: I don't think a warning is necessary here. Thanks for catching these!

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
martinemde: Should we entirely remove the message about disabling the feature and just update the docs?

This would be in addition to adding a command for ignoring a single checksum mismatch, if we decide to do that.

text/0011-gem-checksum-verification.md
martinemde: I might imagine writing a step in CI that runs `bundle lock --frozen` (and I'm making this command up) and the goal would be that it fetches the checksums, but generates an error if anything would be written to the Gemfile.lock.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+
+Path source gems cannot be verified because the checksum of the entire path would be complicated to calculate and unreliable.
+
+### GitHub Dependabot and other automations are at risk of failing.
+
+If a user configures CI to install with frozen bundle, then all dependabot pull requests will fail.
+This will make dependabot update branches much less useful, since a failing update is almost guaranteed.
+In order to remain functional, Dependabot would need to write the corresponding checksum lines to the Gemfile to prevent every CI build from failing.
+This would make dependabot the defacto source of the checksums in the Gemfile instead of rubygems.org or .gem files.
+This might lead users to disable frozen bundles, disable dependabot, or disable the checksums feature, any of which could be considered major drawbacks.
```
deivid-rodriguez, line 177: Dependabot will need updates to work with this feature, and I think that's maybe a good point to make this fully opt-in in the beginning? But I don't understand the point about this leading to users disabling features. As long as Dependabot correctly adds the proper checksums, shouldn't it be fine? That's what Dependabot does for other ecosystems, and it seems fine.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+### How do we handle confusion about the authority of checksums written to the Gemfile.lock
+
+The source of checksums in the Gemfile.lock becomes a matter of trust once it's written.
+Did the checksum come from the API or was it calculated from a .gem file on a developers computer.
+If a checksum error is resolved by one developer in a way that saves an incorrect checksum, how should people know when to approve these changes or not.
+It may not be common practice for most teams to look at the Gemfile.lock when reviewing code.
+Gemfile.lock changes can be hidden in pull request reviews.
+Without a process for checking that the checksums are trustworthy, it's left to every development team to decide on a process.
+
+One solution would be a bundle command that could be run in CI every time the gems are installed that verifies the authenticity of checksums in the Gemfile.lock.
```
deivid-rodriguez, line 217: Technically `bundle install` in frozen mode in CI would catch it, since there won't be a local gem cache?

deivid-rodriguez (commented):
Finally found some time to review this. This is so great @martinemde, impressive! Fully onboard with your careful thorough approach!

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
deivid-rodriguez: What would be the use case for needing this? Could we instead allow users to vendor the specific gem with this checksum? Having random hardcoded checksums in the Gemfile feels weird. Everybody will need to have access to the source `.gem` for it to work, right? So how come would the mismatch happen?

Maybe alternatively (if I understand correctly how this feature would play with vendoring) we could allow users to vendor specific `.gem` files? I think this would be similar to what was requested at https://github.com/rubygems/rubygems/issues/4495?

text/0011-gem-checksum-verification.md
deivid-rodriguez: Would the be fixed in we started recording checksums only for fresh lockfiles? That means people with existing lockfiles (and thus everyone using frozen) would not be affected initially. We could also add a `bundle lock --add-checksums` flag for people to opt-in to the feature on an existing lockfile.

text/0011-gem-checksum-verification.md
martinemde: Ah yes, I could have made it more clear: "initially, dependabot will not be updated and users may disable the feature with the intention of it being temporary, but it could lead to the feature staying disabled for a long time."

text/0011-gem-checksum-verification.md
martinemde: I meant a command for syncing checksums from the remote to ensure you don't have an erroneous checksum generated from a broken gem.

deivid-rodriguez: mmmm, so a command that fixes the issue rather than detecting it, right? Maybe `bundle lock` without a local gem cache including that broken gem should do that, either by default or under an optional flag?

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+# Guide-level explanation
+
+Upon first upgrading to this version of Bundler, expect bundler to continue to be secure by default and to progressively add checksums to the lockfile.
+
+If you wish to immediately add checksums to your lockfile for all locked gems, run `bundle lock`.
+Bundle lock now fetches checksums from remote sources by default.
+If you would like to bypass this behavior, add the `--no-checksums` flag.
+
+Common Bundler commands like `bundle install`, `bundle update`, `bundle lock`, `bundle add` will now automatically record and verify checksums.
+Commands that rely on checksums for verification will output a message or warning when checksums are missing or mismatched.
```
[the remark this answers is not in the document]
segiddins: I can see an argument either way, but it's in my mind a bit of a stretch of what frozen already means (for missing checksums, at least).

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+Writing lockfile to path/to/Gemfile.lock
+$ bundle pristine
+Installing rake 13.0.6
+…
+Installing rspec 3.12.0
+$
+```
+
+When bundler installs a gem from a `.gem` file, it computes the SHA256 checksum of the file.
+If an existing checksum is available from the lockfile or the remote source, it will be compared with the computed checksum at install time.
```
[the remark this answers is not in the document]
segiddins: that makes sense to me. once the checksum is in the lock file, it's checked into source control and should be considered the source of truth for that project

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+
+  To resolve this issue you can either:
+    1. remove the gem at path/to/rake-13.0.6.gem
+    2. run `bundle install`
+  or if you are sure that the new checksum from the lockfile CHECKSUMS at Gemfile.lock:21:17 is correct:
+    1. remove the matching checksum in Gemfile.lock:21:17
+    2. run `bundle install`
+
+To ignore checksum security warnings, disable checksum validation with
+  `bundle config set --local disable_checksum_validation true`
```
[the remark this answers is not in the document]
segiddins: IMO no, if we want a way to avoid putting checksums in the lock file (which im also not sure we want), it should be under a different flag

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+This could be a larger burden for some bundles on slower machines.
+
+Future proposals could attempt to address this slowdown, if necessary.
+
+### Unverifiable gems
+
+When gems come from private gem servers that do not implement the compact index, checksums will not be available.
+Bundler will calculate digests from .gem files from sources that don’t supply checksums.
+
+An additional flag for `bundle lock` could be provided that allows reinstalling, and thus calculating the checksum for, gems that don't have a checksum.
```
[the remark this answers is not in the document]
segiddins: > So the behavior for remote servers will be that there's only one level of verification, i.e., that the checksum of whatever gem that was once installed is recored in the lockfile and should be the same in for future installs.

it will also be checked against what the server reports in the compact index, if that is present (which is what already happens)

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+
+Gems from a git source are verified by nature of matching a specific git SHA and should be excluded from checksum verification.
+
+Path source gems cannot be verified because the checksum of the entire path would be complicated to calculate and unreliable.
+
+### GitHub Dependabot and other automations are at risk of failing.
+
+If a user configures CI to install with frozen bundle, then all dependabot pull requests will fail.
```
segiddins, line 173: why will they fail? will dependabot not add the new lock file checksum in when it runs `bundle install`?

segiddins: if not, please add a small section with instructions for how automation should be updated to be compatible

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
deivid-rodriguez: I'm also not sure we want it, but I reckon it could be useful for a graceful migration, given that the setting already exists, even if we'll eventually remove the flag.

It'd be interesting to check why we added this flag in the first place.

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
deivid-rodriguez: In the missing case, I meant only for GEM sources (which if I understood correctly are the only ones adding checksum). And also only if the Gemfile is already using CHECKSUMS.

text/0011-gem-checksum-verification.md
deivid-rodriguez: Dependabot (unfortunately) runs a fixed Bundler version. So until that is updated to a version that knows about checksums, it won't add checksums.

On top of that, Dependabot (also unfortunately) uses Bundler internals a lot, rather than only shelling out to Bundler commands, so it's likely that stuff breaks.

deivid-rodriguez:
Overall my main concern here would be how to roll this out.

I tend to think the safest way to battle test this before making it the default would be:
* Add checksums only to fresh lockfiles, and let existing lockfiles work as they work now for the moment.
* Additionally, provide an `--add-checksums` flag to `bundle lock` for users to opt-in to the feature.

This would be consistent with how I'm approaching https://github.com/rubygems/rubygems/pull/5700, by applying it only to new lockfiles, and having already a `bundle lock --add-platform` command to opt-in to locking more platforms in an existing lockfile.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+Using checksums, sourced from rubygems.org or from digests of local .gem file, is a more strict way of locking to a specific gem version.
+The feature should work transparently in much the same way that bundler already locks to specific versions by ensuring that not just the same version, but the exact same rubygem file data, is installed on each environment.
+
+# Guide-level explanation
+
+Upon first upgrading to this version of Bundler, expect bundler to continue to be secure by default and to progressively add checksums to the lockfile.
+
+If you wish to immediately add checksums to your lockfile for all locked gems, run `bundle lock`.
+Bundle lock now fetches checksums from remote sources by default.
+If you would like to bypass this behavior, add the `--no-checksums` flag.
```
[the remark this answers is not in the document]
martinemde: This is enabled by default and every install or update will save and compare checksums.

However, for most people they are not performing a clean install. If you run bundle install and nothing is installed from gem, you get no checksums. I wanted to provide a way for people to explicitly dive in fully.

I think `bundle lock` should just always add all available checksums from all sources to the lock.

The full explicit record and test currently works via `bundle lock --update --bundler` (to avoid gem changes) and then `bundle pristine`. I want `lock` to grab checksums without the extra args.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+
+# Motivation
+
+Verifying gem checksums with known checksums at install time is a stronger way to verify that the exact same gem source is being used in every environment where the bundle is installed.
+Using checksums, sourced from rubygems.org or from digests of local .gem file, is a more strict way of locking to a specific gem version.
+The feature should work transparently in much the same way that bundler already locks to specific versions by ensuring that not just the same version, but the exact same rubygem file data, is installed on each environment.
+
+# Guide-level explanation
+
+Upon first upgrading to this version of Bundler, expect bundler to continue to be secure by default and to progressively add checksums to the lockfile.
```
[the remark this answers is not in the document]
martinemde: I kept stumbling on my own words here. I could remove this line and the RFC would be just fine.

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
martinemde: When a checksum is loaded for a gem, from anywhere, we check if another checksum is already in memory. If so, we merge them (which means comparing, raising on mismatch, or combining on match).

On any given run you usually compare either an API and a lockfile checksum or a lockfile and a gem digest checksum.

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
martinemde: Good question. It was there when I accidentally stole this work from @segiddins. It's a failsafe in case we really mess someone up or they like using hacked and corrupted gems and wish to continue doing so.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+
+Future compatibility for new checksum algorithms is supported by reading and writing existing checksums for installed gems even if the algorithm is unknown.
+Checksum comparisons take into account the algorithm used and raises errors accordingly.
+
+Bundler commands that interact with checksums either fetch checksums from sources and update the Gemfile.lock or compare checksums in the Gemfile.lock with the computed digest of the gem being installed.
+
+Internal storage of checksums indexes the checksums by a gem's NameTuple (name, version, platform) and the checksum's algorithm.
+The source of each checksum is stored with the checksum so that errors can describe how to fix conflicting checksums.
+
+When two remote sources have a checksum for the same gem, the are compared.
```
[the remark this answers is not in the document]
martinemde: Ah yes. Funny story:

When you fixed the code so multi-source gemfiles don't fall back to full index if they all support APIs, it started making a bunch of tests fail for the checksums branch.

All the example tests for multi source gemfiles produce gems with different checksums even when the tests intend for them to be the same. It directly conflicts with the other way we combine checksums when building the index.

My choice here was to either make an ugly patch that disables checksums in that situation, or try to support it.

I added support and made a good error message that says, basically, "your sources don't agree about the checksums of this gem, we can't fix it locally and you'll need to update your sources in order to continue"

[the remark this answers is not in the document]
martinemde: Message test here: https://github.com/rubygems/rubygems/blob/master/bundler/spec/install/gemfile/sources_spec.rb

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
martinemde: On first install, perhaps a `bundle pristine` if desired, we will compute, compare and store all checksums for all .gem files. They will end up with a checksum this way if they were missing one.

As for an explicit way to fill in missing gems besides that, I don't know.


I'm actually concerned about another similar case, a broken gem that someone depends on and needs. Theoretically the server sends a bad checksum and the gem continually clashes. Can you ignore just that gem?

Also, missing checksums that keep breaking frozen bundles, like for different platforms or that didn't get recorded. Frozen says "raises if lock would be changed."

text/0011-gem-checksum-verification.md
martinemde: Now that I think about it, I think it should sync and compare everything. This is the chance to print that "the gem you have installed generated a checksum that is different than what the server is saying."

Bundle lock should just pull all checksums every time then? Lock previously doesn't hit the network unless a resolve is necessary. This would make it always hit the network unless `--no-checksums` or `--local` is set.

martinemde: Ah, my explicit concern is this scenario:

1. Malicious developer sets checksum and upload gem.
2. Checksum would be invalid if compared with server, but doesn't happen during install.
3. Malicious gem is successfully installed.

If we had a process that could verify your checksums against the server without altering your bundle, it could run in CI as part of tests and help protect against malicious or bad checksums in lockfiles.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,217 @@
...
+
+Example:
+
+```
+$ bundle pristine
+Installing rake 13.0.6
+…
+Installing rspec 3.12.0
+42 gems without checksums.
+Use `bundle lock` to add checksums to Gemfile.lock.
```
martinemde, line 52: I originally wrote this because it seemed like a nice interface, but now I wonder how I can reliably tell how many gems _could_ have checksums that don't.

Is this useful and we should add it?

martinemde: This also brings up the problem of not continually printing this when the source doesn't support checksums. If the source doesn't support it, do we add a special checksum like `unavailable=true` and then use that as a marker for a gem that doesn't need to be verified? We do support just about any key value pair in the checksum field.

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
martinemde: Bundler docs:
> Do not allow the Gemfile.lock to be updated after this install. Exits non-zero if there are going to be changes to the Gemfile.lock.

> frozen (BUNDLE_FROZEN): Disallow changes to the Gemfile. When the Gemfile is changed and the lockfile has not been updated, running Bundler commands will be blocked. Defaults to true when --deployment is used.

I think there's 2 ways we could interpret this:

1. The first quote would suggest that we cannot make any change to the Gemfile.lock, so a frozen bundle should fail if checksums are missing. People use frozen for added security (don't install something different than what was committed and tested) and this fits with added security. If your production environment wants to install the alpine linux version of a bundled gem and you don't have a checksum for it then hard fail.

2. The second quote seems to say that if there are changes to the _Gemfile_ not reflected in the Gemfile.lock, then fail. A missing checksum for a bundle that would otherwise work in frozen mode is not a clear violation of that rule. With this approach, bundler does not become more secure, and installing the alpine linux version of a gem without being able to verify the checksum is ok.

Option 1 seems implied to me, but option 2 is fully within existing expectations and will break less CI builds. The question is: _should we break those CI builds?_ Is that what those engineers expected bundler to do? If frozen doesn't break that build, then what will?

martinemde:
@deivid-rodriguez for the most part this works so smoothly and invisibly that many people may not notice it.

However, for the frozen bundle situation there are some drawbacks. My instinct is that we should fail in that case, but we hopefully have provided a clear way to avoid the failure in the way of messaging how to add all checksums.

Almost all failure will trigger in CI on the "update bundler version" commit. Then everyone will become aware of how frozen impacts the deployment. Then they will run bundle lock, commit again, and they should be good to go.

Please check my assumptions if you disagree.

text/0011-gem-checksum-verification.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Example CHECKSUM section
+
+A freshly created Rails 7.0.7.2 app creates the following CHECKSUMS section (snipped for brevity):
+
+```
+CHECKSUMS
+  actioncable (7.0.7.2) sha256=a921830a59ee314939955c9fc3b922d2b1f3ebc16fdf062370b9078aa0dc28c5
+  actionmailbox (7.0.7.2) sha256=33aeae209fc876c072e5ad28c7ffc16ace533d391368ad6390bb6183c2b27a24
+  actionmailer (7.0.7.2) sha256=0e9061159af8c220042b7714a2ba01e2d71d2904f308021ec714793e5f9811a0
```
martinemde, line 125: An idea that might be a little to late:

```suggestion
  sha256:
    a921830a59ee314939955c9fc3b922d2b1f3ebc16fdf062370b9078aa0dc28c5 actioncable (7.0.7.2)
    33aeae209fc876c072e5ad28c7ffc16ace533d391368ad6390bb6183c2b27a24 actionmailbox (7.0.7.2)
    0e9061159af8c220042b7714a2ba01e2d71d2904f308021ec714793e5f9811a0 actionmailer (7.0.7.2)
```

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
deivid-rodriguez: My interpretation of frozen has always been 1. For example, if we run under a new platform that's not already in the lockfile, we fail in frozen mode even if there's no Gemfile changes. Basically in frozen mode, everything should be taken from the lockfile.

In addition to this, I think this feature kind of misses the point if we don't fail when we find a mismatch and instead happily install the mismatched gem.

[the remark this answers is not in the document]
deivid-rodriguez: > If frozen doesn't break that build, then what will?

Exactly.

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
deivid-rodriguez: Nice 👍.

text/0011-gem-checksum-verification.md
deivid-rodriguez: I see. I think this is indeed a risk and another reason to let this feature be battle tested for a while in an opt-in only mode first. I don't understand the hurry to force everyone into checksums before we know the full implications.

text/0011-gem-checksum-verification.md
deivid-rodriguez: I think it's fine for `bundle lock` to hit the network (or local cache) to look for checksums.

I'm not sure I understand your concern, can you elaborate this 3 steps? What does "sets checksum" mean? Manually edit a lockfile changing the checksum of a gem? What does "upload gem" mean in this context? Are you implying a situation where a malicious actor has both control of your lockfile, and your remote server? That sounds wild and I don't see how anything can be done in that situation 🤣

Overall, I think `bundle install --force` in frozen mode should be a strict way to check remote checksums against lockfile checksums, since that bypasses the local cache and should not allow lockfile changes.

text/0011-gem-checksum-verification.md
deivid-rodriguez: When I first saw this, I understood `bundle install` by default would only advertise checksums, but not add them, and I thought that was ideal, regardless of showing a number or not. I.e., let people know that this feature is available and tell them how to enable it. And don't change behavior unless a CHECKSUM section is present in the lockfile.

deivid-rodriguez:
I think not being sure whether we want to even raise if we find mismatches in frozen mode is a symptom that this should be battle tested with users first for a period.

In my opinion, this feature, when enabled, should be strict. But it should not get in the middle unless something bad is happening. I think there are many edge cases that we may only find once this starts being used by people.

Hence my take of letting people experiment with this first, listen to feedback, fix issues, and only then start enforcing checksums.

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
martinemde: I think you're right. Anything broken like that should be a fail and we should address the failure from it, not preemptively prepare for it.

> So checksums won't play nice with these lockfiles I believe. And they are still the majority.

This seems like the distinction between frozen (strict) and unfrozen. If you want to use bundler loosely, then you also won't get this extra assurance from checksums.

text/0011-gem-checksum-verification.md
martinemde: Opt-in vs opt-out is interesting. We'll get a lot more useful (if potentially unhappy) feedback from opt-out, but maybe we need an opt-in period first?

Opt in to me would be triggered by running `bundle lock --checksums" and it would add the checksums. We would consider the feature enabled if the lockfile either didn't exist or already had checksums.

My experience using the feature in testing is that most of the time the checksums are not added right away. I think only the frozen-on-ci situation will cause problems with opt-out, and it will hopefully be found at the same time as the person upgrades bundler, so it will be intentional.

I'm open to either really.

text/0011-gem-checksum-verification.md
martinemde: hehe, maybe my imagination is running a bit wild. With that access you can do whatever you want.

The presence of bundle install --force should provide options for people wanting to use this feature like that.

The more we talk about it, the more clear it is that we can't anticipate everything.

text/0011-gem-checksum-verification.md
martinemde: Given that we really can't anticipate the problems of this feature, I'm starting to lean this way.

So the plan would be to build into this feature a line that suggests running `bundle lock --checksums` at the end of every install message until the CHECKSUMS exist in the Gemfile.lock. Until then, we treat the feature as disabled.

text/0011-gem-checksum-verification.md
deivid-rodriguez: Yes, that would be my preferred approach. You get checksums when there's no prior lockfile, or when you explicitly add them.

If a lockfile has no checksums, they can only be added explicitly.

text/0011-gem-checksum-verification.md
deivid-rodriguez: > With that access you can do whatever you want.

Exactly, we shouldn't care about it since at that point there's no much we can do.

text/0011-gem-checksum-verification.md
deivid-rodriguez: Yes! I think a line in `bundle install` output is not too invasive and properly advertises the feature.

text/0011-gem-checksum-verification.md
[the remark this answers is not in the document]
deivid-rodriguez: Agreed, if a lockfile is using this loose mode, then you don't want checksums.

If you do explicitly add checksums to that, that means you opt-in to strict mode, and you'll only get gems without specific platforms locked in there, with their respective checksums, and that's what will be consistently installed, regardless of the platform where the lockfile is used.

PR #52, opened 2023-10-01
PostgreSQL major upgrade "meta" proposal.

simi:
[rendered proposal](https://github.com/rubygems/rfcs/blob/pg-upgrade/text/0000-postgres-major-upgrade.md/)

[phase 1 RFC](https://github.com/rubygems/rfcs/pull/53) ([rendered proposal](https://github.com/rubygems/rfcs/blob/pg-upgrade-1/text/0000-postgres-major-upgrade-1.md/))

_reference scripts will be available soon_

### demo (pardon poor playback quality please)

[Záznam obrazovky z 2023-09-30 23-02-39.webm](https://github.com/rubygems/rfcs/assets/193936/a5f98716-c971-43f5-b6e0-7344ec8bff18)

PR #53, opened 2023-10-01
PostgreSQL upgrade Phase 1 proposal.

simi:
[rendered proposal](https://github.com/rubygems/rfcs/blob/pg-upgrade-1/text/0000-postgres-major-upgrade-1.md/)

see first [meta RFC](https://github.com/rubygems/rfcs/pull/52) for more context


indirect (approved):
This looks good. Maybe we can upgrade to `db.m7g.xlarge`, since it seems like the RDS monitors show this instance working well, and newer instance types are more efficient?

I would suggest a pre-announced window for the restart on Saturday or Sunday to disrupt existing users as little as possible. Hopefully the caching system will keep fetching and downloading gems working even while the DB restarts, but I guess we'll be finding out whether that works in real life as well as in theory.

text/0000-postgres-major-upgrade-1.md
```diff
@@ -0,0 +1,56 @@
...
+
+# Motivation
+
+The key goal of this phase is to ensure that all crucial tools, configurations, and scripts are in place, facilitating a seamless migration from PostgreSQL 11 to 12. By making the most of the mandatory downtime for enabling replication, we also include the "pg_stat_statements" for enhanced monitoring, maximizing the value derived from this period.
+
+# Guide-level explanation
+
+The initial phase of the PostgreSQL upgrade will include:
+
+1. Setting up `pgbouncer` in the Kubernetes cluster with associated services to manage connections.
```
segiddins, line 18: will it be set up in the `rubygems.org` repo, and then in the per-environment namespaces? or globally in terraform?

segiddins (approved):
👍🏻 to everything andre said as well

text/0000-postgres-major-upgrade-1.md
simi: IMHO it is enough to run a "maintenance" pod within cluster with `pgbouncer` installed and use it for migration. It will be needed only for few minutes for migration switch as a proxy holding connections opened.

simi:
:information_source: updated with reference scripts link and `db.m7g.xlarge`. DB restart already has happened with no problems (it was matter of 10s downtime).

PR #57, opened 2024-10-24
Gem Push Sigstore Bundles RFC

segiddins:
Signed-off-by: Samuel Giddins <segiddins@segiddins.me>

### DIFF
--- a/text/0000-rubygems.org-sigstore-bundles.md
+++ b/text/0000-rubygems.org-sigstore-bundles.md
@@ -1,0 +1,57 @@
+- Feature Name: RubyGems.org accepts Sigstore Bundles with Gem Pushes
+- Start Date: 2024-10-24
+- RFC PR: (leave this empty)
+- Bundler Issue: (leave this empty)
+
+# Summary
+
+RubyGems.org will add a new variant to the gem push endpoint that allows gems authors to include sigstore bundles with their gem pushes.
+RubyGems.org will verify that the sigstore bundle is valid and signed by a trusted publisher configured for the gem.
+
+# Motivation
+
+This is one of the first steps to bringing signed cryptographic attestations of gem provenance to the Ruby ecosystem.
+Allowing gem authors to attach provenance information to the artifacts they publish is a prerequisite for downstream consumers to be able to verify the provenanve of the gems they install.
+
+# Guide-level explanation
+
+[Sigstore](https://sigstore.dev) is a suite of projects that aims to make signing & verifying artifacts easy and secure.
+Sigstore bundles are a way to package verification materials (signing certificate, proof of inclusion in a tamper-evident trust log, signed time source) along with information about an artifact. In other words, a sigstore bundle will allow you to verify "GitHub repo segiddins/mygem at commit 1234567 with github action workflow 'release.yml' attests to mygem-1.0.0.gem having sha256 hash abcd1234".
+There will be a new way of uploading gems, along with an array of such attestations that RubyGems.org will verify and store.
+
+# Reference-level explanation
+
+## `POST /api/v1/gems`
+
+This is the existing endpoint for gem pushes.
+Currently, rubygems.org assumes that the entire body of the request is the gem to be processed.
+
+We will update the endpoint to behave slightly differently when the `Content-Type` header is set to `multipart/mixed` or `multipart/form-data`.
+In that case:
+
+The request MUST include a part with name `gem` that contains the gem file. The `gem` part SHOULD have a `Content-Type` header of `application/octet-stream`.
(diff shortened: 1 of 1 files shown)


simi (approved):

text/0000-rubygems.org-sigstore-bundles.md
```diff
@@ -0,0 +1,57 @@
...
+## `POST /api/v1/gems`
+
+This is the existing endpoint for gem pushes.
+Currently, rubygems.org assumes that the entire body of the request is the gem to be processed.
+
+We will update the endpoint to behave slightly differently when the `Content-Type` header is set to `multipart/mixed` or `multipart/form-data`.
+In that case:
+
+The request MUST include a part with name `gem` that contains the gem file. The `gem` part SHOULD have a `Content-Type` header of `application/octet-stream`.
+The request MAY include a part with name `attestations` that contains sigstore bundles. The `sigstore` part MUST have a `Content-Type` header of `application/json` and MUST be a valid JSON array of sigstore bundle object.
```
haydentherapper, line 33: Will you be opinionated on the contents of the bundles? Namely, should the bundle only contain attestations, or will you accept raw signatures over artifacts?

text/0000-rubygems.org-sigstore-bundles.md
```diff
@@ -0,0 +1,57 @@
+- Feature Name: RubyGems.org accepts Sigstore Bundles with Gem Pushes
+- Start Date: 2024-10-24
+- RFC PR: (leave this empty)
+- Bundler Issue: (leave this empty)
+
+# Summary
+
+RubyGems.org will add a new variant to the gem push endpoint that allows gems authors to include sigstore bundles with their gem pushes.
+RubyGems.org will verify that the sigstore bundle is valid and signed by a trusted publisher configured for the gem.
```
haydentherapper, line 9: To confirm, will trusted publishing be a requirement to accept Sigstore bundles, as in you won't accept signatures generated by a non-CI workflow identity?

text/0000-rubygems.org-sigstore-bundles.md
segiddins: Hmmm. Right now I have it accepting what sigstore-ruby can verify against an artifact. That includes hashedrekord entries, if that's your question

text/0000-rubygems.org-sigstore-bundles.md
segiddins: To start, I plan on limiting to trusted publishing, since it gives us a solid basis for an identity to verify against. I believe this is the approach that pypi is taking. Going forward though, I see no reason why rubygems.org fundamentally needs to restrict what attestations we accept, given they are coming from an actor that has permissions to publish to the gem

haydentherapper: One benefit for only accepting trusted publishing identities is if you want to provide a default verification policy, as you don't have to worry about exposing non-public maintainer details (except for dealing with private repos).

text/0000-rubygems.org-sigstore-bundles.md
segiddins: Hmm this is a good question that actually dovetails with what clients should be doing when verifying...

segiddins: I think I'm going to leave things as-is for now, and we can always restrict what attestations we allow publishing with in the future.

As-is, both raw signatures and DSSE envelopes will verify, so long as the dsse envelope is an in-toto statement with a subject that matches the artifact

text/0000-rubygems.org-sigstore-bundles.md
```diff
@@ -0,0 +1,57 @@
...
+RubyGems.org will verify that any provided attestations are valid, match the uploaded gem, and signed by a trusted publisher configured for the gem.
+
+## `GET /api/v1/attestations/:full_name.json`
+
+This is a new endpoint that will allow users to download the attestations for a given gem.
+This endpoint will return a JSON array of sigstore bundle objects.
+
+# Drawbacks
+
+This is adding a second variant to a critical endpoint, which could increase complexity and maintenance burden.
```
deivid-rodriguez, line 44: Did we consider alternatives that preserve the current endpoint, like:

* Use a completely different endpoint, maybe `POST /api/v1/attestations/:full_name.json`, to push gems with attestations?
* Use 2 separate requests, one for pushing the gem like we do now, one for pushing the attestations.

Not sure if those would make sense at all, but mentioning just in case.

segiddins: Ah yeah I did consider both options

We want the attestations pushed atomically with the gem, otherwise it becomes possible for clients to not see the attestations in the window between the POSTs

Additionally, keeping the same endpoint means we don't need to pick a non-RESTy way of representing gem pushes

